PR #53, opened 2020-05-09
Add zip bombs and test cases

gfs:

PR #56, opened 2020-05-09
Rewrite Zip extractor for hybrid enumerable

### LINKED ISSUE #55: Hybrid Parallel-Enumerable Extractor
Try to gain some of the benefits of parallelization without having to load the entire extracted contents of the archive into memory before returning.

## Create an internal queue to push to. 
This continually fills up the stack until the file is fully extracted, but memory is freed as the caller pulls off the enumerable.
Allows callers to start pulling entries immedietly instead of waiting for all processing to be done.
Frees up memory as objects are yielded.
This subverts the expectations of what an enumerable does. It will continue processing between yields.

```csharp
ZipFile zipFile = new ZipFile(fileEntry.Content);
var zipEntries = new List<ZipEntry>();
foreach (ZipEntry zipEntry in zipFile)
{
    zipEntries.Add(zipEntry);
}
ConcurrentStack<FileEntry> files;
bool done = false;
_ = Task.Run(() => {
    Parallel.ForEach(zipEntries, zipEntry => 
    {
        using var memoryStream = new MemoryStream();
        byte[] buffer = new byte[BUFFER_SIZE];
        var zipStream = zipFile.GetInputStream(zipEntry);
        StreamUtils.Copy(zipStream, memoryStream, buffer);    
        var newFileEntry = new FileEntry(zipEntry.Name, fileEntry.FullPath, memoryStream);
        files.PushRange(ExtractFile(newFileEntry, true))
    });
    done = true;
});
while (!done){
    if (files.TryPop(out FileEntry result)){
        yield return result;
    }
    else{
        Thread.Sleep(1);
    }
}
```
## Process by chunks
Do parallel processing on a chunk of files at a time, stopping in between chunks to yield until empty.

```csharp
ZipFile zipFile = new ZipFile(fileEntry.Content);
var zipEntries = new List<ZipEntry>();
foreach (ZipEntry zipEntry in zipFile)
{
    zipEntries.Add(zipEntry);
}
ConcurrentStack<FileEntry> files;
while (zipEntries.Count > 0)
{
    if (zipEntries.Count > MIN_BATCH_SIZE){
        int batchSize = Math.min(MAX_BATCH_SIZE,zipEntries.Count);
        Parallel.ForEach(zipEntries[0..batchSize], zipEntry => 
        {
            using var memoryStream = new MemoryStream();
            byte[] buffer = new byte[BUFFER_SIZE];
            var zipStream = zipFile.GetInputStream(zipEntry);
            StreamUtils.Copy(zipStream, memoryStream, buffer);    
            var newFileEntry = new FileEntry(zipEntry.Name, fileEntry.FullPath, memoryStream);
            files.PushRange(ExtractFile(newFileEntry, true))
        });
        zipEntries.RemoveRange(0, batchSize);
    }
    else{
        foreach (var zipEntry in zipEntries){
            using var memoryStream = new MemoryStream();
            byte[] buffer = new byte[BUFFER_SIZE];
            var zipStream = zipFile.GetInputStream(zipEntry);
            StreamUtils.Copy(zipStream, memoryStream, buffer);    
            var newFileEntry = new FileEntry(zipEntry.Name, fileEntry.FullPath, memoryStream);
            foreach (var extractedFile in ExtractFile(newFileEntry, parallel))
            {
                yield return extractedFile;
            }        
        }
        zipEntries.Clear();
    }
    while (files.TryPop(out FileEntry entry))
    {
        yield return entry;
    }
}
```

### LINKED ISSUE #57: Use FileStreams instead of Memory Streams
Arrays in .NET have a limit of 2 billion elements in a single element array.

 Calling .ToArray() like [here](https://github.com/microsoft/OSSGadget/blob/92e1552c36f0c3c11c132b918196ad0b30ab4011/src/MultiExtractor/GnuArFile.cs#L31) on a memory stream longer than 2GB should (?) thus cause only some of the bytes to be returned (or perhaps an error).

I think we should rewrite this to just parse bytewise off of the end of the memory stream. reading individual chunks into buffers as needed.

### LINKED ISSUE #58: Support Greater than ~2GB file entries
Doing this requires rewriting the parser to create a multidimensional array of bytes because individual dimensions are limited to 2 billion. https://docs.microsoft.com/en-us/dotnet/api/system.array.length?view=netcore-3.1

gfs:
Fix #55 - Rewrite parallelization to yield enumerables for more responsive behavior.
Fix #57 - Support Lookup table based .ar files (need examples for test cases)
Fix #58 - Support > 32 bit sized files using FileStream instead of MemoryStream
Carry through parallel setting to sub-extractions + sub-extractions of initially non-parallelizable archives. Supercede #52 
Add Quine and ZipBomb tests. Supercede #53


gfs:
I can't reproduce the test failure locally.  All tests pass on my machine.  @scovetta can you check out and see if the tests run for you?  

Assuming its passing locally for you, I've had trouble with this .rar file before specifically in the pipeline.  Should we just remove that test case?

gfs:
Tests passing locally for me. Only the parallel nested test is failing, and only on the pipeline.

PR #59, opened 2020-05-11
User/jacobmsft/download cache

jacobmsft:
Add download caching options to the relevant tools


src/Shared/PackageManagers/CargoProjectManager.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using AngleSharp.Css;
```
gfs, line 4: ?

src/oss-characteristics/CharacteristicTool.cs
```diff
@@ -100,33 +103,33 @@ public CharacteristicTool()
...
 
-            string targetDirectoryName = null;
-            while (targetDirectoryName == null || Directory.Exists(targetDirectoryName))
+            var downloadTool = new DownloadTool();
+            // ensure that the cache directory has the required package, download it otherwise
+            bool cached = !string.IsNullOrEmpty(targetDirectoryName);
+            if (cached)
             {
-                targetDirectoryName = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+                directoryNames.AddRange(await downloadTool.EnsureDownloadExists(purl, targetDirectoryName));
```
gfs, line 118: I don't understand this.  It looks like you went from always adding the result of download tool to directoryNames to only doing it if you are cached.  But if you're not cached directoryNames would always be an empty list.

src/oss-characteristics/CharacteristicTool.cs
```diff
@@ -100,33 +103,33 @@ public CharacteristicTool()
...
             }
-
-            Logger.Trace("Creating directory [{0}]", targetDirectoryName);
-            Directory.CreateDirectory(targetDirectoryName);
-
-            var downloadTool = new DownloadTool();
-            var directoryNames = await downloadTool.Download(purl, targetDirectoryName);
```
gfs, line 119: If you are not cached it looks like you no longer ever try to download

src/oss-download/DownloadTool.cs
```diff
@@ -5,8 +5,10 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
```
gfs, line 8: I don't see what requires this in this file.

src/tests/DownloadTests.cs
```diff
@@ -134,7 +134,7 @@ private async Task TestDownload(string purl, string targetFilename)
             {
                 try
                 {
-                    downloadTool.Download(packageUrl, tempDirectoryName).Wait();
+                    downloadTool.EnsureDownloadExists(packageUrl, tempDirectoryName).Wait();
```
gfs, line 137: You should test the `ensureDownloadExists(url,null)` case as well

src/oss-characteristics/CharacteristicTool.cs
jacobmsft: That is right, there is a bug...I made multiple iterations and missed the loop. I've abstracted all the package downloading capabilities in a new class PackageDownloader. This should be fixed now, thanks

src/oss-characteristics/CharacteristicTool.cs
jacobmsft: fixed

src/tests/DownloadTests.cs
jacobmsft: added

src/oss-download/DownloadTool.cs
jacobmsft: there were unwanted usings in various files...removed

src/oss-defog/DefoggerTool.cs
```diff
@@ -159,8 +160,8 @@ public async Task AnalyzePackage(PackageURL purl)
         {
             Logger.Trace("AnalyzePackage({0})", purl.ToString());
 
-            var downloadTool = new DownloadTool();
-            foreach (var directory in await downloadTool.Download(purl))
+            var packageDownloader = new PackageDownloader(purl, (string)(this.Options["cache-directory"]));
```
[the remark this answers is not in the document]
gfs: Actually why does AnalyzePackage have two constructors?

Wouldn't `AnalyzePackage(PacakageURL purl, string? directory = null)` and then pass through directory to `PackageDownloader` work just as well?

That would also get rid of this case where you implicitly check the cache directory.

src/oss-detect-cryptography/DetectCryptographyTool.cs
```diff
@@ -232,23 +229,14 @@ public DetectCryptographyTool()
...
-            {
-                targetDirectoryName = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
-            }
-
-            Logger.Trace("Creating directory [{0}]", targetDirectoryName);
-            Directory.CreateDirectory(targetDirectoryName);
-
-            var downloadTool = new DownloadTool();
-            var directoryNames = await downloadTool.Download(purl, targetDirectoryName);
+            var packageDownloader = new PackageDownloader(purl, (string)(this.Options["cache-directory"]));
```
gfs, line 238: Bug? You don't use the `targetDirectoryName` argument and instead pull `cache-directory` out of options.

src/oss-detect-cryptography/DetectCryptographyTool.cs
```diff
@@ -277,10 +265,11 @@ public async Task<List<IssueRecord>> AnalyzePackage(PackageURL purl)
             {
                 Logger.Warn("Error downloading {0}.", purl.ToString());
             }
-            
+
```
gfs, line 268: Whitespace

src/oss-download/PackageDownloader.cs
```diff
@@ -0,0 +1,258 @@
...
+            if (purl == default)
+            {
+                Logger.Warn("Invalid PackageURL (null)");
+                return new List<string>();
+            }
+
+            List<string> downloadDirectories = new List<string>();
+            foreach (var version in this.PackageVersions)
+            {
+                downloadDirectories.AddRange(await this.Download(version, metadataOnly, doExtract, this.doCache));
```
gfs, line 173: You could run a linter to remove unneeded `this` calls

src/oss-download/PackageDownloader.cs
```diff
@@ -0,0 +1,258 @@
...
+            if (purl == null)
+            {
+                throw new ArgumentNullException("purl cannot be null");
+            }
+
+            this.destinationDirectory = destinationDir;
+            // if we are told to use a cache dir, and it exists, believe that caching is still doable
+            if (string.IsNullOrEmpty(destinationDir))
+            {
+                this.doCache = false;
```
gfs, line 51: I am not a fan of this pattern.  I think you should explicitly tell the package downloader if downloads should use cache or not - it shouldn't be based on the content of the destinationDirectory argument.

src/tests/DownloadTests.cs
```diff
@@ -20,135 +20,182 @@ public class DownloadTests
...
+            var downloadTool = new DownloadTool(); // do common initialization
+            PackageDownloader packageDownloader1 = await DownloadPackage(packageUrl, tempDirectoryName);
+            string dir = Directory.GetCurrentDirectory();
+            var fileMatchCount = Directory.EnumerateFiles(tempDirectoryName, targetFilename, SearchOption.AllDirectories).Count();
+            Assert.IsTrue(fileMatchCount > 0);
+
+            // do that again, this should not do anything since the cache already has the package
+            PackageDownloader packageDownloader2 = await DownloadPackage(packageUrl, tempDirectoryName);
+            fileMatchCount = Directory.EnumerateFiles(tempDirectoryName, targetFilename, SearchOption.AllDirectories).Count();
+            Assert.IsTrue(fileMatchCount > 0);
```
gfs, line 153:
```suggestion
            Assert.IsTrue(Directory.EnumerateFiles(tempDirectoryName, targetFilename, SearchOption.AllDirectories).Any());
```

It might not matter here, but in general `Any()` is faster than `Count() > 0`.

src/tests/ExtractorTests.cs
```diff
@@ -131,6 +131,7 @@ public void TestQuineBombs(string fileName, bool parallel)
             {
                 return;
             }
+            // Getting here means we didnt catch the bomb
```
gfs, line 134: Remove this

src/Shared/PackageManagers/CargoProjectManager.cs
```diff
@@ -37,22 +37,30 @@ public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl,
...
             }
 
+            var url = $"{ENV_CARGO_ENDPOINT}/api/v1/crates/{packageName}/{packageVersion}/download";
             try
             {
-                var url = $"{ENV_CARGO_ENDPOINT}/api/v1/crates/{packageName}/{packageVersion}/download";
+                string targetName = $"cargo-{purl.ToStringFilename()}";
+                string extractionPath = Path.Combine(TopLevelExtractionDirectory, targetName);
+                // if the cache is already present, no need to extract
+                if (doExtract && Directory.Exists(extractionPath) && cached == true)
```
gfs, line 46: Minor improvement.  Shortcuts evaluation of Directory.Exists if cached isn't set.
```suggestion
                if (doExtract && cached && Directory.Exists(extractionPath))
```

src/oss-download/DownloadTool.cs
```diff
@@ -55,8 +53,16 @@ static async Task Main(string[] args)
                 {
                     try
                     {
+                        bool.TryParse(downloadTool.Options["extract"]?.ToString(), out bool doExtract);
+                        bool.TryParse(downloadTool.Options["download-metadata-only"]?.ToString(), out bool metadataOnly);
+                        string targetDirectory = (string)downloadTool.Options["download-directory"];
+                        // are we caching? 
+                        bool doCaching = !string.IsNullOrEmpty(targetDirectory);
```
gfs, line 60: I might have missed this on the last one.  Looks like another cache where caching is implicit depending on if target directory is set.  What if you want the tool to download something in a download folder thats not "." but you don't want cache?

src/oss-risk-calculator/RiskCalculatorTool.cs
```diff
@@ -53,14 +51,15 @@ static void Main(string[] args)
 
             if (((IList<string>)riskCalculator.Options["target"]).Count > 0)
             {
-                Dictionary<string, int> freq = new Dictionary<string, int>();
+                string destinationDirectory = (string)riskCalculator.Options["cache-directory"];
+                bool doCaching = string.IsNullOrEmpty(destinationDirectory);
```
gfs, line 55: This looks like the opposite logic of what you normally do where caching is set based on destinationDirectory being not null.

jacobmsft: I should stop late night coding

jacobmsft:
> I see one bug candidate in `RiskCalculatorTool.cs`, `DoCache` is usually `!= string.IsNullOrEmpty` but here it is `= string.IsNullOrEmpty`.
> 
> It would be better to have argument to cache be distinct from the argument for directory in all cases - see comment in `DownloadTool.cs` in particular, but it looks like at least `RiskCalculatorTool.cs` has the same problem.
> 
> If you don't want to address that in this PR can you create follow-up issues and link them?

yes, that can be a separate PR, if that requirement needs to be added to the code. For my requirement, I just need the destinationDirectory but the functionality is separate in the PackageDownloader class - its just a matter of adding a flag.

src/oss-defog/DefoggerTool.cs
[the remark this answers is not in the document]
jacobmsft: I guess they are separate functions because the tool might need the ability to analyze a directory that is not a package - not sure what is the right change here

jacobmsft:
Thanks for all the review iterations Gabe,

src/oss-defog/DefoggerTool.cs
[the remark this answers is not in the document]
gfs: Pushed a commit that fixes this so the constructor matches the others with the directory and cache options.

gfs:
I did a little tinkering.  I fixed one bug and rewrote the caching in DownloadTool the way I'd suggested.  No need for the follow-up PR.

jacobmsft:
Since we are doing the downloadTool --use-cache arguments here in this PR, I'm adding some more changes to keep the options consistent with the tools, and fixing some bugs from previous checkin

gfs:
Sounds good. I didn't do it for all tools because many specify a cache directory and for those it's safe to use that as an indicator of if you should cache but for download tool the argument was the target not necessarily a cache.

gfs:
Checked out the changes you made. I like it. 👍

jacobmsft:
Thanks again for spending much time reviewing it, and sorry for some of the bugs that came up because of constant refactoring ...next time should be better :)
Merging to master now.

PR #62, opened 2020-05-15
Add support for ISO, VHD, VHDX to MultiExtractor

gfs:
Fix #61

Adds support for single + multi-threaded parsing of ISO, VHD and VHDX files.

PR #64, opened 2020-05-15
Debugging Test Failures

gfs:
Experimental PR to run pipelines.  Trying to nail down what is causing test failures in #62


gfs:
It looks like the failure is caused by the changes in minimagic.cs

gfs:
Missing stream reset in minimagic.  Fixed in #62

PR #73, opened 2020-05-17
Add WIM and VMDK support and BSD ar

### LINKED ISSUE #67: Add more disc formatting support to VHD/VHDX reading
DiscUtils looks like it supports:

https://github.com/DiscUtils/DiscUtils/blob/b3ffb5123eea8791a90d6a2beb65c68cc5708b24/Library/DiscUtils.FileSystems/DiscUtils.FileSystems.csproj#L11-L18
Btrfs
HfsPlus (worth doing alongside #63) 
SquashFs
Xfs

in addition to the EXT, FAT and NTFS support we already have.

### LINKED ISSUE #74: Expanded .ar format support
https://github.com/xoofx/LibObjectFile

Supports BSD, GNU and DEB formats. Doesn't appear to have a way to distinguish directly from the stream, looks like you need to know which kind it is.  Will need to implement a checker either in LibObjectFile library (probably better) or in MiniMagic.

We have an .ar of each type checked into the `TestData` folder:
Common: Like Deb, already available in test data as `Shared.deb`
GNU: Available in test data as `Shared.ar`
BSD: Available in test data as `Shared.bsd.ar`

gfs:
Supercedes #72 
Fix #67
Fix #74


src/tests/ExtractorTests.cs
```diff
@@ -111,9 +122,8 @@ public void TestMiniMagic(string fileName, ArchiveFileType expectedArchiveFileTy
         [DataRow("zblg.zip", true)]
         [DataRow("zbsm.zip", false)]
         [DataRow("zbsm.zip", true)]
-        //[DataRow("zbxl.zip", false)]
+        //[DataRow("zbxl.zip", false)] //These work but have stalled the anti-malware scanner on the pipeline
```
scovetta, line 125: Maybe they need our zip bomb detection code?

gfs: Interestingly it doesn't stall running defender locally.

PR #96, opened 2020-05-31
Separate pipelines for OSS Gadget and Multiextractor

### LINKED ISSUE #94: Only publish new versions of MultiExtractor to NuGet if MultiExtractor has changed.
# Problem
Currently, OSS Gadget Characteristic depends on Application Inspector which depends on MultiExtractor.

Here's what happens when we update anything in OSS Gadget:

1. On every update to any OSS Gadget component we publish a new version of all the OSS Gadget tools including Multiextractor.

2. To be up to date, we then need to update the reference in Application Inspector for MultiExtractor.  This publishes a new version of Application Inspector.

3. We then need to update OSS Gadget Characteristics with the reference to the new version of Application Inspector.  Go to 1.

# Proposed Solution
Only publish a new version of each tool to NuGet if it in particular has changed. In the above example, instead of publishing a new version of MultiExtractor, only Characteristic tool has changed and we skip the dependency loop.

## Possible implementation
This could be done with separate sections for each OSS Gadget tool in the pipeline.

It may still be desirable to publish the full set to GitHub, but that adds some complexity.  It may be fine to only publish the tools that have changed.

## Additional Benefits
Less version churn (especially no-op version upgrades) for users who are only referencing one library or tool - for example just MultiExtractor.

gfs:
Fix #94

OSSGadget build code based on https://mattvsts.github.io/2019/05/04/solving-the-looping-problem-in-Azure-DevOps-Pipelines/


Pipelines/core-pipeline.yml
```diff
@@ -43,7 +45,7 @@ stages:
     - task: DotNetCoreCLI@2
       displayName: Dotnet Test
       inputs:
-        command: 'test'
+        command: 'test --filter ClassName=Microsoft.CST.OpenSource.Tests.DetectCryptographyTests.cs|ClassName=Microsoft.CST.OpenSource.Tests.DownloadTests.cs|ClassName=Microsoft.CST.OpenSource.Tests.FindSourceTests.cs|ClassName=Microsoft.CST.OpenSource.Tests.VersionTests.cs'
```
scovetta, line 48: Is there a better way of doing this that doesn't involve hardcoding the test names here?

gfs:
Putting the multi extractor tests in a separate project would work

Pipelines/core-pipeline.yml
gfs: Fixed.

PR #100, opened 2020-06-02
Update DiscUtils to MyGet version

gfs:


jacobmsft:
checkin after fixing the tests of course

daalcant:
I can get the restore working with the NuGet.Config file, but the DiscUtils version (0.16.0-alpha0020) being targeted isn't available at the MyGet feed selected (yet). Looks like the DiscUtils CI deployment is failing: 
https://ci.appveyor.com/project/LordMike/discutils

This PR is blocked by the above.

gfs:
I swear this works locally. I wonder why it's not in the feed from the server. I'll take another look tomorrow.

daalcant:
> I swear this works locally. I wonder why it's not in the feed from the server. I'll take another look tomorrow.

I may be wrong about the package not being available, but I can't work out the errors I'm seeing, for example:

``NU1605: Detected package downgrade: DiscUtils.Btrfs from 0.16.0-alpha0020 to 0.16.0-alpha0017. Reference the package directly from the project to select a different version. 
 MultiExtractor.Tests -> Microsoft.CST.MultiExtractor -> DiscUtils 0.16.0-alpha0020 -> DiscUtils.Btrfs (>= 0.16.0-alpha0020) 
 MultiExtractor.Tests -> DiscUtils.Btrfs (>= 0.15.1-ci0002)``

Initially I thought the downgrade was due to unavailability of the 0.16.0-alpha0020 package, but after a bit more research it looks like a dependency resolution conflict.
https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution#nearest-wins

Problem is, I don't see where the package reference for ``DiscUtils.Btrfs (>= 0.15.1-ci0002)`` is defined anywhere within MultiExtractor.Test.csproj.

gfs:
That version happens to be the old version I was using and the newest version on nuget. Those might be hints.

daalcant:
Working to isolate a potential fix for this from a test branch

daalcant:
@gfs This appears to be in good shape now. I'll work the package renaming updates into another branch.

PR #104, opened 2020-06-04
Refactor and Rename Extractor

daalcant:
Renames MultiExtractor to RecursiveExtractor. This will create a new NuGet package as well.

One more PR will be required after this one to complete the change to OSSGadget using RecursiveExtractor NuGet package references, and update the CI pipeline.


daalcant:
@gfs @jacobmsft @scovetta PR checks passed, but I would appreciate any assistance with spot checking local builds.

jacobmsft:
> 
> 
> @gfs @jacobmsft @scovetta PR checks passed, but I would appreciate any assistance with spot checking local builds.

Do I need to add a new nuget source for building this? Right now the latest pre-release package I can see for Discutils is 0.15.1-ci0002

gfs:
The correct nuget.config was set up for getting DiscUtils from MyGet in a separate PR. It is in the multi extractor folder.

daalcant:
We can move the Nuget.Config into the src/ folder next to the solution if that's cleaner. The original config file was only for the Extractor project, so I left it there, but the Nuget Restore function checks for configs all the way up to the repo root.

jacobmsft:
> 
> 
> The correct nuget.config was set up for getting DiscUtils from MyGet in a separate PR. It is in the multi extractor folder.

That worked, thanks! Just curious, are we using some features in the myget discutils that are not in the nuget.org one, and will this nuget.config eventually go away?

jacobmsft:
> 
> 
> We can move the Nuget.Config into the src/ folder next to the solution if that's cleaner. The original config file was only for the Extractor project, so I left it there, but the Nuget Restore function checks for configs all the way up to the repo root.
> 
> Currently you can restore with:
> `cd src`
> `nuget restore -ConfigFile RecursiveExtractor/NuGet.Config`

Is the build pipeline working by using the custom nuget.config now? If this is a permanent thing, then might make sense to move it into the solution folder, since VS expects it there for the solution nuget restore - (there is a  PR for the build pipelines for different tools to be split, will that work if we move the nuget config to solution level?) 
If the multiextractor project is going to remain in the OSSGadget solution, we could move it up, provided that the build pipeline works. IMO, since the MultiExtractor has a nuget and (going to) have a separate build definition, maybe it can have its own solution file, and doesn't have to be loaded up with OSSGadget (OSSGadget can consume the nuget). What do you think?

gfs:
There's a fix I contributed that is in the 0.16 series but 0.15 hasn't been updated for a year. It looks like they use MyGet now instead of Nuget so it's unlikely we will be able to get rid of the nuget.config

daalcant:
> 
> 
> > We can move the Nuget.Config into the src/ folder next to the solution if that's cleaner. The original config file was only for the Extractor project, so I left it there, but the Nuget Restore function checks for configs all the way up to the repo root.
> > Currently you can restore with:
> > `cd src`
> > `nuget restore -ConfigFile RecursiveExtractor/NuGet.Config`
> 
> Is the build pipeline working by using the custom nuget.config now? If this is a permanent thing, then might make sense to move it into the solution folder, since VS expects it there for the solution nuget restore - (there is a PR for the build pipelines for different tools to be split, will that work if we move the nuget config to solution level?)
> If the multiextractor project is going to remain in the OSSGadget solution, we could move it up, provided that the build pipeline works. IMO, since the MultiExtractor has a nuget and (going to) have a separate build definition, maybe it can have its own solution file, and doesn't have to be loaded up with OSSGadget (OSSGadget can consume the nuget). What do you think?

True, it would be more intuitive to move the config to allow restoring from the solution. I'll move it into src/ and merge this tonight.

jacobmsft:
perfect!

PR #116, opened 2020-06-11
Fixes initial FileEntry creation to have correct path

gfs:
Run CodeMaid


gfs:
Requesting review for CodeMaid formatting.

Settings used are checked in at `CodeMaid.Config` in the root directory.

jacobmsft:
I'm assuming these are all formatting changes in the csharp files and approve it.

gfs:
This is the only real code change - a fix to remove a double print of the base archives name in RecursiveExtractor

https://github.com/microsoft/OSSGadget/blob/95751cd2f77609619de616f972b8d684b18695de/src/RecursiveExtractor/Extractor.cs#L173

jacobmsft:
> This is the only real code change - a fix to remove a double print of the base archives name in RecursiveExtractor
> 
> https://github.com/microsoft/OSSGadget/blob/95751cd2f77609619de616f972b8d684b18695de/src/RecursiveExtractor/Extractor.cs#L173

LGTM, also, is there a rule for CodeMaid to limit lines to 80 chars or less? That might help readability

scovetta:
We should align on a standard CodeMaid config (haven't used it before) -- but also not sure that we should go with 80 as our limit -- maybe 110? [Slashdot article](https://linux.slashdot.org/story/20/05/31/211211/linus-torvalds-argues-against-80-column-line-length-coding-style-as-linux-kernel-deprecates-it).

gfs:
The default in code maid is wrapping comments (not lines) at 100.  I'm not sure there's a setting to format code lines.  I can bump it to 110.

gfs:
I bumped the setting to 110 and removed regions because I ended up not liking what it did - it wasn't rearranging things into the regions on save.

You can get CodeMaid here: https://marketplace.visualstudio.com/items?itemName=SteveCadwallader.CodeMaid

gfs:
I am not in favor of wrapping all lines (vs comments), but we should try to stick to ~120 characters when possible.

gfs:
@jacobmsft I went through Extractor.cs and wrapped lines closer to 120 per your comment manually.

jacobmsft:
> 
> 
> @jacobmsft I went through Extractor.cs and wrapped lines closer to 120 per your comment manually.

Sounds good...its not a must do for me, but it was helpful when Resharper used to do it in VS (normal VS word wrap is not as helpful), Makes online code-viewing/reviewing easier. Thought codemaid also might have it...nevermind.

PR #136, opened 2020-07-27
Support netstandard2.0 in Recursive Extractor

gfs:


gfs:
@daalcant can you investigate the pipeline failures?

gfs:
The problem is that the pipeline tries to publish RecursiveExtractor, however RecursiveExtractor is multitargetting netstandard2.0 and 2.1 and publish then requires a target framework argument.  Instead RecursiveExtractor should be `dotnet pack`ed which will include both versions.

PR #157, opened 2020-10-01
Retarget all projects to .net 5.0 rc

jacobmsft:
.net 5 is production ready.

### DIFF
--- a/Pipelines/core-pipeline.yml
+++ b/Pipelines/core-pipeline.yml
@@ -28,7 +28,6 @@ stages:
   jobs:
   - template: templates/dotnet-test-job.yml
     parameters:
-      dotnetVersion: 3.1.x
       projectPath: 'src/oss-tests/oss-tests.csproj'
 
 - stage: SDL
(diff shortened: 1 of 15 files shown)


gfs:
@jacobmsft You also will need to update the pipeline to use .NET 5.  

Here's an example snippet
```
steps:
- task: UseDotNet@2
  displayName: 'Install .NET Core SDK'
  inputs:
    version: 5.0.x
    performMultiLevelLookup: true
    includePreviewVersions: true # Required for preview versions
```

from https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/dotnet-core?view=azure-devops

gfs (changes requested):
This needs the pipeline work alongside it.

jacobmsft:
> 
> 
> This needs the pipeline work alongside it.

good point

gfs:
@daalcant can you double check my pipeline work?

daalcant:
@gfs Looks mostly fine to me. I'm not a fan of "Use Pre-Release .NET SDK" defaulting to true. IMO should default to false and be set via the respective parameter. If the team is fine with it then I am too though.

gfs:
Half a dozen of this six of another. I'm fine changing it to be the other way it'll just mean we have to update core pipeline instead of the templates when 5.0 is official released. Either way we need to update it in multiple places.

scovetta:
.NET Core 5 should be GA pretty soon, so I think it's ok to jump on the latest release. We're only telling developers they need to use a release candidate if they want to compile OSS Gadget - the platform-specific releases should work fine.

But I also don't feel super strong about waiting until GA to make this switch.

daalcant:
I'm in favor of moving forward without changes then.

gfs (approved):

PR #171, opened 2020-11-24
Update CodeQL Action with .NET 5

daalcant:
Update CodeQL Action with a step to install the .NET 5 SDK


gfs:
This looks like an nbgv error.

You can try to update the dependency anywhere its referenced and also update to the latest version in `directory.build.props`.

daalcant:
@gfs Thanks for the assist. Looks like the update fixed the set-env issues and the codeql workflow is succeeding again.

PR #182, opened 2020-12-08
Fix #179 (dynamic NuGet API endpoints)

### LINKED ISSUE #179: Fix NuGet API
We hardcode the wrong NuGet API endpoint:

Always go to https://api.nuget.org/v3/index.json and get the current RegistrationsBaseUrl.

scovetta:
This change updates the NuGet manager to dynamically find the right endpoint, which should fix some missing packages, as well as minor fixes to a few other files.

### DIFF
--- a/src/Shared/PackageManagers/NuGetProjectManager.cs
+++ b/src/Shared/PackageManagers/NuGetProjectManager.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Linq;
 using System.Text.Json;
 using System.Threading.Tasks;
+using System.Xml.Linq;
 
 namespace Microsoft.CST.OpenSource.Shared
 {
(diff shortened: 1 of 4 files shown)


src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -7,16 +7,66 @@
...
+                        Logger.Warn(ex, "Error parsing NuGet API endpoint: {0}", ex.Message);
+                    }
+                }
+            }
+            catch(Exception ex)
+            {
+                Logger.Warn(ex, "Error parsing NuGet API endpoint: {0}", ex.Message);
+            }
+            RegistrationEndpoint = NUGET_DEFAULT_REGISTRATION_ENDPOINT;
+            return NUGET_DEFAULT_REGISTRATION_ENDPOINT;
```
jacobmsft, line 69: shouldn't it be return RegistrationEndpoint;?

scovetta: Shouldn't matter - they're the same value.

gfs: I recommend changing it in case the previous line changes.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -162,11 +306,15 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
                     if (!string.IsNullOrEmpty(repoCandidate))
                     {
-                        PackageURL repoPurl = GitHubProjectManager.ExtractGitHubPackageURLs(repoCandidate).ToList().FirstOrDefault();
-                        mapping.Add(repoPurl, 1.0F);
-                        return mapping;
+                        var candidate = ExtractGitHubPackageURLs(repoCandidate).FirstOrDefault();
+                        if (candidate != null)
+                        {
+                            mapping.Add((PackageURL)candidate, 1.0F);
+
```
gfs, line 313: Whitespace

gfs (approved):
Broadly LGTM.

src/Shared/PackageManagers/NuGetProjectManager.cs
scovetta: Thanks! Fixed.

src/Shared/PackageManagers/NuGetProjectManager.cs
scovetta: @gfs Would `mapping.Add(candidate as PackageURL, 1.0F);` work too? Is there a different because a c-style cast and the `as` operator?

gfs: I think (but haven't tested) that that would work as well if you like how it looks.

gfs: The difference appears to be:

* as operator: failure to cast at runtime returns null. Compile time checks for types. 
* Casting: failure to cast at runtime throws an exception. Compile time checks for types. 

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast

PR #185, opened 2021-01-06
Adds Binary and Compressed detection

gfs:


src/oss-defog/DefoggerTool.cs
```diff
@@ -97,6 +222,113 @@ public EncodedString(EncodedStringType Type, string Filename, string EncodedText
...
+                {
+                    totalBytesRead += bytesRead;
+
+                    if (totalBytesRead == readBuffer.Length)
+                    {
+                        int nextByte = stream.ReadByte();
+                        if (nextByte != -1)
+                        {
+                            byte[] temp = new byte[readBuffer.Length * 2];
+                            Buffer.BlockCopy(readBuffer, 0, temp, 0, readBuffer.Length);
```
jacobmsft, line 307: What I understand here is that you are expanding the readBuffer based on the contents of the stream. Is there a way to store the length of the stream in the respective classes? That way, you can do just one allocation and just keep adding bytes to it without using the temp array.

src/oss-defog/DefoggerTool.cs
```diff
@@ -373,6 +392,24 @@ private static void Main(string[] args)
...
                         }
+
+                        var blobDir = (string?)defoggerTool.Options["save-blobs-to"];
+                        var blobNumber = 0;
+                        foreach (var blobFinding in defoggerTool.NonTextFindings)
+                        {
+                            Logger.Info("Blob: {0}: {1}", blobFinding.Filename, blobFinding.EncodedText);
+                            if (blobDir is string)
+                            {
+                                var path = Path.Combine(blobDir, blobFinding.Filename, $"archive-{archiveNumber++}");
```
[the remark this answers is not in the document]
gfs: Good catch.

src/oss-defog/DefoggerTool.cs
```diff
@@ -674,6 +724,10 @@ private void ParseOptions(string[] args)
                         Options["save-archives-to "] = args[++i];
                         break;
 
+                    case "--save-blobs-to":
+                        Options["save-blobs-to "] = args[++i];
```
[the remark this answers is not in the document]
gfs: Thanks.

src/oss-defog/DefoggerTool.cs
```diff
@@ -54,28 +55,152 @@ internal class DefoggerTool : OSSGadget
...
+            JAVA
+        }
+
+        public static byte[] HexStringToBytes(string hex)
+        {
+            try
+            {
+                if (hex is null) { throw new ArgumentNullException(nameof(hex)); }
+
+                var ascii = new byte[hex.Length / 2];
```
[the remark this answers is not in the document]
gfs: I'll take a closer look to see if I can optimize it a bit.

src/oss-defog/DefoggerTool.cs
gfs: We just have Stream which doesn't have a length property. You need to read a stream to its end to find its length.  If this were a string stream we could just use a StreamReader.ReadToEnd(), but its a binary stream so there's no convenient equivalent class I'm aware of.  If there's a more clever way that doesn't use a temp array that's fine but I don't think you can do it with one allocation.

If I rewrote some stuff I could use something like a MemoryStream instead of Stream possibly.

jacobmsft: hmmm...yea I suspected that might be the reason, feel free decide what to do based on if it could significantly improve performance.

src/oss-defog/DefoggerTool.cs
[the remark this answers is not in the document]
gfs: I rewrote it in LINQ with no bit shifts or try block.  I found a very esoteric way to do it on stack overflow but I think LINQ is the most readable.  Here's the complex way.  What do you think? https://stackoverflow.com/a/9995303

src/oss-defog/DefoggerTool.cs
gfs: I rewrote this entirely without any temp allocations.  Take another look.

jacobmsft: This looks awesome and much cleaner! :)

src/oss-defog/DefoggerTool.cs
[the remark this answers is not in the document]
jacobmsft: I was checking this same post too yesterday :) LINQ is pretty maintainable, so unless this code is a hotpath, LGTM.

src/oss-defog/DefoggerTool.cs
gfs: I looked closer at this and I don't think I need this at all.  Removed the whole thing.

jacobmsft (approved):

src/oss-defog/DefoggerTool.cs
```diff
@@ -54,28 +55,131 @@ internal class DefoggerTool : OSSGadget
         /// </summary>
         private const int INTERESTING_STRINGS_CUTOFF = 24;
 
+        /// <summary>
+        ///     Enum of executable types
+        /// </summary>
+        public enum EXECUTABLE_TYPE
```
scovetta, line 61: Any particular reason why these are ALL_CAPS instead of PascalCase?

src/oss-defog/DefoggerTool.cs
```diff
@@ -130,7 +289,53 @@ private static void Main(string[] args)
...
+                        var archiveDir = (string?)defoggerTool.Options["save-archives-to"];
+                        var archiveNumber = 0;
+                        foreach (var archiveFinding in defoggerTool.ArchiveFindings)
+                        {
+                            Logger.Info("[Archive] {0}: {1} -> {2}", archiveFinding.Filename, archiveFinding.EncodedText, archiveFinding.Type);
+                            if (archiveDir is string)
+                            {
+                                Directory.CreateDirectory(archiveDir);
+                                var path = Path.Combine(archiveDir, archiveFinding.Filename, $"archive-{archiveNumber++}");
+                                Logger.Info("Saving to ", path);
```
scovetta, line 320: I think this needs to be:
```
Logger.Info("Saving to: {0}", path);
```

src/oss-defog/DefoggerTool.cs
```diff
@@ -130,7 +289,53 @@ private static void Main(string[] args)
...
+                        }
+
+                        var blobDir = (string?)defoggerTool.Options["save-blobs-to"];
+                        var blobNumber = 0;
+                        foreach (var blobFinding in defoggerTool.NonTextFindings)
+                        {
+                            Logger.Info("[Blob] {0}: {1}", blobFinding.Filename, blobFinding.EncodedText);
+                            if (blobDir is string)
+                            {
+                                var path = Path.Combine(blobDir, blobFinding.Filename, $"archive-{archiveNumber++}");
```
scovetta, line 334: Should this be archiveNumber of blobNumber?

src/oss-defog/DefoggerTool.cs
gfs: Nope. I copied them over from another project that uses that format for enums. I can change them.

src/oss-defog/DefoggerTool.cs
gfs: Fixed.

src/oss-defog/DefoggerTool.cs
gfs: Fixed.

src/oss-defog/DefoggerTool.cs
```diff
@@ -213,6 +421,16 @@ public void AnalyzeFile(string filename)
             var fileContents = File.ReadAllText(filename);
 #pragma warning restore SEC0116 // Path Tampering Unvalidated File Path
 
+            AnalyzeFile(filename, fileContents);
+        }
+
+        public bool HasNonTextContent(string content)
+        {
+            return content.Any(ch => char.IsControl(ch) && ch != '\t' && ch != '\r' && ch != '\n');
```
scovetta, line 429: Might be unnecessary, but [Char.IsWhiteSpace](https://docs.microsoft.com/en-us/dotnet/api/system.char.iswhitespace?view=net-5.0) handles unicode whitespace too.

src/oss-defog/DefoggerTool.cs
gfs: Fixed.

src/oss-defog/DefoggerTool.cs
gfs: I think that should work.  I was being more restrictive because whitespace might mean like vertical tab which doesn't make sense in the context of code.  I'll just use the built in instead though.

scovetta (approved):

0x53414D:
lgtm

PR #194, opened 2021-02-22
Fixes 193 and 192

### LINKED ISSUE #192: Consider a sleep delay for oss-find-squats
Since the typosquatting detector is making a sizeable number of calls to different endpoints, it would be useful to have a default sleep delay between each call (maybe 250 ms by default, but settable with `--sleep-delay MSECS`). I think this is important, especially for the Ubuntu endpoints, since IIRC, I've been throttled by hitting the search endpoint too frequently.

### LINKED ISSUE #193: Increase suffixes for oss-find-squats
Suggest adding a few more suffixes (1, 4, 5, 6, 7) here -- I've seen them in the wild a bit.

Generative.cs:

```
var suffixes = new string[] { "s", "2", "3", "ng", "-ng", ".", "x", "-", "_", "js" };
```

gfs:
Fix #192 
Fix #193


scovetta (approved):
LGTM

PR #201, opened 2021-02-25
Add oss-diff

gfs:
Adds the oss-diff tool for diffing packages.  Respects NO_COLOR. Has arguments for Before lines After lines and Overall Context, Output Format and Crawling Archives.

### DIFF
--- a/src/OSSGadget.sln
+++ b/src/OSSGadget.sln
@@ -30,7 +30,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "oss-tests", "oss-tests\oss-
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "oss-metadata", "oss-metadata\oss-metadata.csproj", "{7E9A489A-B502-4622-A34C-EF507E18DFEB}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "oss-find-squats", "oss-find-squats\oss-find-squats.csproj", "{23D17E08-9E95-4448-B75D-BABAF1F3C854}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "oss-find-squats", "oss-find-squats\oss-find-squats.csproj", "{23D17E08-9E95-4448-B75D-BABAF1F3C854}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "oss-diff", "oss-diff\oss-diff.csproj", "{17C9F321-8A49-4140-B009-AD476F959D42}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -85,6 +87,10 @@ Global
 		{23D17E08-9E95-4448-B75D-BABAF1F3C854}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{23D17E08-9E95-4448-B75D-BABAF1F3C854}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{23D17E08-9E95-4448-B75D-BABAF1F3C854}.Release|Any CPU.Build.0 = Release|Any CPU
+		{17C9F321-8A49-4140-B009-AD476F959D42}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{17C9F321-8A49-4140-B009-AD476F959D42}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{17C9F321-8A49-4140-B009-AD476F959D42}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{17C9F321-8A49-4140-B009-AD476F959D42}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/src/oss-diff/DiffTool.cs
+++ b/src/oss-diff/DiffTool.cs
@@ -38,7 +38,7 @@ namespace Microsoft.CST.OpenSource
             public bool UseCache { get; set; }
 
             [Option('d', "delete-after-diff", Required = false, Default = false,
-                HelpText = "delete the packages after diffing them.")]
+                HelpText = "Delete the packages after diffing them.")]
             public bool DeleteAfterDiff { get; set; }
 
             [Option('B', "context-before", Required = false, Default = 0,
@@ -169,6 +169,18 @@ namespace Microsoft.CST.OpenSource
 
                                 break;
                         }
+                    }
+                }
+
+                if (options.DeleteAfterDiff)
+                {
+                    foreach(var directory in locations)
+                    {
+                        Directory.Delete(directory, true);
+                    }
+                    foreach (var directory in locations2)
+                    {
+                        Directory.Delete(directory, true);
                     }
                 }
             }
--- a/src/oss-diff/DiffTool.cs
+++ b/src/oss-diff/DiffTool.cs
@@ -204,6 +204,7 @@ namespace Microsoft.CST.OpenSource
                         }
                     }
 
+                    diffObjs.Sort((x,y) => x.startLine1.CompareTo(y.startLine1));
                     foreach (var diff in diffObjs)
                     {
                         var sb = new StringBuilder();
--- a/src/oss-diff/DiffTool.cs
+++ b/src/oss-diff/DiffTool.cs
@@ -53,6 +53,14 @@ namespace Microsoft.CST.OpenSource
                 HelpText = "Number of lines to give as context. Overwrites Before and After options.")]
             public int Context { get; set; } = 0;
 
+            [Option('a', "added-only", Required = false, Default = false,
+                HelpText = "Only show added lines (and requested context).")]
+            public bool AddedOnly { get; set; } = false;
+
+            [Option('r', "removed-only", Required = false, Default = false,
+                HelpText = "Only show removed lines (and requested context).")]
+            public bool RemovedOnly { get; set; } = false;
+
             [Value(0, Required = true,
                 HelpText = "PackgeURL(s) specifier to analyze (required, repeats OK)", Hidden = true)] // capture all targets to analyze
             public IEnumerable<string> Targets { get; set; } = Array.Empty<string>();
@@ -130,28 +138,34 @@ namespace Microsoft.CST.OpenSource
                         switch (line.Type)
                         {
                             case ChangeType.Inserted:
-                                if (beforeBuffer.Any())
+                                if (!options.RemovedOnly || options.AddedOnly)
                                 {
-                                    foreach(var buffered in beforeBuffer)
+                                    if (beforeBuffer.Any())
                                     {
-                                        Console.WriteLine($"  {buffered.Pastel(Color.Gray)}");
+                                        foreach (var buffered in beforeBuffer)
+                                        {
+                                            Console.WriteLine($"  {buffered.Pastel(Color.Gray)}");
+                                        }
+                                        beforeBuffer.Clear();
                                     }
-                                    beforeBuffer.Clear();
+                                    afterCount = options.After;
+                                    Console.WriteLine($"+ {line.Text}".Pastel(Color.Green));
                                 }
-                                afterCount = options.After;
-                                Console.WriteLine($"+ {line.Text}".Pastel(Color.Green));
                                 break;
                             case ChangeType.Deleted:
-                                if (beforeBuffer.Any())
+                                if (!options.AddedOnly || options.RemovedOnly)
                                 {
-                                    foreach (var buffered in beforeBuffer)
+                                    if (beforeBuffer.Any())
                                     {
-                                        Console.WriteLine($"  {buffered.Pastel(Color.Gray)}");
+                                        foreach (var buffered in beforeBuffer)
+                                        {
+                                            Console.WriteLine($"  {buffered.Pastel(Color.Gray)}");
+                                        }
+                                        beforeBuffer.Clear();
                                     }
-                                    beforeBuffer.Clear();
+                                    afterCount = options.After;
+                                    Console.WriteLine($"- {line.Text}".Pastel(Color.Red));
                                 }
-                                afterCount = options.After;
-                                Console.WriteLine($"- {line.Text}".Pastel(Color.Red));
                                 break;
                             default:
                                 if (afterCount-- > 0)
--- a/src/oss-diff/DiffTool.cs
+++ b/src/oss-diff/DiffTool.cs
@@ -98,26 +98,37 @@ namespace Microsoft.CST.OpenSource.DiffTool
                 Logger.Error($"Format {options.Format} is not supported.");
                 return string.Empty;
             }
-            PackageURL purl1 = new PackageURL(options.Targets.First());
-            PackageURL purl2 = new PackageURL(options.Targets.Last());
-            var manager = ProjectManagerFactory.CreateProjectManager(purl1, options.DownloadDirectory ?? Path.GetTempPath());
-            var manager2 = ProjectManagerFactory.CreateProjectManager(purl2, options.DownloadDirectory ?? Path.GetTempPath());
+
 
             // Map relative location in package to actual location on disk
             Dictionary<string, (string, string)> files = new Dictionary<string, (string, string)>();
             IEnumerable<string> locations = Array.Empty<string>();
             IEnumerable<string> locations2 = Array.Empty<string>();
 
-            if (manager is not null)
+            try
             {
-                locations = await manager.DownloadVersion(purl1, true, options.UseCache);
+                PackageURL purl1 = new PackageURL(options.Targets.First());
+                var manager = ProjectManagerFactory.CreateProjectManager(purl1, options.DownloadDirectory ?? Path.GetTempPath());
+
+                if (manager is not null)
+                {
+                    locations = await manager.DownloadVersion(purl1, true, options.UseCache);
+                }
             }
-            else
+            catch(Exception)
             {
                 var tmpDir = Path.GetTempFileName();
                 File.Delete(tmpDir);
-                extractor.ExtractToDirectory(tmpDir, options.Targets.First());
-                locations = new string[] { tmpDir };
+                try
+                {
+                    extractor.ExtractToDirectory(tmpDir, options.Targets.First());
+                    locations = new string[] { tmpDir };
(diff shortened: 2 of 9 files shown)


src/oss-diff/oss-diff.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
+    <RootNamespace>oss_diff</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Shared\Shared.csproj" />
```
scovetta, line 10: Missing DiffPlex reference.

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,128 @@
...
+            
+        }
+        static async Task Main(string[] args)
+        {
+            var diffTool = new DiffTool();
+            await diffTool.ParseOptions<Options>(args).WithParsedAsync<Options>(diffTool.RunAsync);
+        }
+
+        public async Task RunAsync(Options options)
+        {
```
scovetta, line 55: Crash when passed no arguments:

```
Unhandled exception. System.InvalidCastException: Unable to cast object of type 'System.Runtime.CompilerServices.NullableAttribute' to type 'CommandLine.Text.UsageAttribute'.
```

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,128 @@
...
+                        case ChangeType.Inserted:
+                            Console.ForegroundColor = ConsoleColor.Green;
+                            Console.Write("+ ");
+                            break;
+                        case ChangeType.Deleted:
+                            Console.ForegroundColor = ConsoleColor.Red;
+                            Console.Write("- ");
+                            break;
+                        default:
+                            Console.ForegroundColor = ConsoleColor.Gray; // compromise for dark or light background
```
scovetta, line 116: You'll probably want to limit the context to N  (default 3?) lines before/after a change so the output isn't huge. 

Also, not sure if .NET handles this automatically, but if the environment variable NO_COLOR is set, just output plain text.

gfs:
The color here was lifted from diffplex example. I can just remove it.

I don’t think this includes any context the way I’ve written it.

scovetta:
> The color here was lifted from diffplex example. I can just remove it. I don’t think this includes any context the way I’ve written it.

Actually I like the coloring, it just sometimes messes up output when you redirect to a file.

And you're right, no context -- but in that case, maybe we should add a little :-)

gfs:
Got it. I’ll add those options for context size (default 0) and color (default on)

src/oss-diff/oss-diff.csproj
gfs: This somehow didn't get pushed. Fixed.

src/oss-diff/DiffTool.cs
gfs: Good catch.

src/oss-diff/DiffTool.cs
gfs: I'll try switching to one of the libraries listed here like Pastel that auto support NO_COLOR

https://no-color.org

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,207 @@
...
+                            files[key] = existing;
+                        }
+                        else
+                        {
+                            files[key] = (string.Empty, contents);
+                        }
+                    }
+                }
+
+                foreach (var filePair in files)
```
jacobmsft, line 130: Ill create some feature requests for this, including recognizing directory offsetting, recursive extraction of content packages, json/sarif support etc. (for example, if color output is run in a container (toolshed), the output will come out garbled.

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,207 @@
...
+                                        beforeBuffer.RemoveAt(0);
+                                    }
+                                }
+
+                                break;
+                        }
+                    }
+                }
+
+                if (options.DeleteAfterDiff)
```
jacobmsft, line 193: Is this flag kind of the same as caching? Allowing caching means the packages wont be deleted after download, in addition to not downloading if present in the first place - that's the behavior in other tools, might make to just keep it consistent.

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            if (options.Targets.Count() != 2)
+            {
+                Logger.Error("Must provide exactly two packages to diff.");
+                return;
+            }
+
+            if (options.Context > 0)
+            {
+                options.Before = options.Context;
```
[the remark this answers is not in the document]
gfs: Can you clarify what you mean without indexing?

src/oss-diff/DiffTool.cs
gfs: The library I'm using respects NO_COLOR already per Mike's request.

src/oss-diff/DiffTool.cs
gfs: Fair point. I'll use the cache argument instead.

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,189 @@
...
+                foreach (var directory in locations)
+                {
+                    foreach (var file in Directory.EnumerateFiles(directory, "*", SearchOption.AllDirectories))
+                    {
+                        var contents = File.ReadAllText(file);
+                        files.Add(string.Join(Path.DirectorySeparatorChar,file.Substring(directory.Length).Split(Path.DirectorySeparatorChar)[2..]), (contents, string.Empty));
+                    }
+                }
+
+                foreach (var directory in locations2)
```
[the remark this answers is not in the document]
gfs: Fixed.

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,189 @@
...
+            }
+
+            if (options.Context > 0)
+            {
+                options.Before = options.Context;
+                options.After = options.Context;
+            }
+
+            (PackageURL purl1, PackageURL purl2) = (new PackageURL(options.Targets.First()), new PackageURL(options.Targets.Last()));
+            var manager = ProjectManagerFactory.CreateProjectManager(purl1, options.DownloadDirectory ?? Path.GetTempPath());
```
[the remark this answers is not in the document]
gfs: The only one that appears to use a static variable is VSM Project Manager.  I'm not sure if that one was you or @scovetta or the motivation behind using the static cache in it.

src/oss-diff/DiffTool.cs
gfs: But for the other ones definitely create some feature requests

src/oss-diff/DiffTool.cs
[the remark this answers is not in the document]
jacobmsft: nevermind, I got it wrong...you are just creating two copies of the Context object.

src/oss-diff/DiffTool.cs
gfs: I added sarif and file location output options.

What do you mean by directory offsetting?

Do you want me to run all the found files through recursive extractor?

gfs: Recursive extractor bit ended up easier than I anticipated. Now included here.

scovetta (commented):
This is off to a great start -- couple recommendations:

1. I think the default experience should be either to extract archives, or at least to show the context within corresponding files of the two packages. For example, when running `.\oss-diff.exe pkg:npm/left-pad@1.2.0 pkg:npm/left-pad@1.3.0`, I think I'd expect to see something like the output of `diff -r DIR1 DIR2`:

```
miscovet@DESKTOP-KDQF1J0:/mnt/c/dev/source/OSSGadget/src/oss-diff/bin/Debug/net5.0$ diff -r npm-left-pad@1.2.0/npm-left-pad@1.2/ npm-left-pad@1.3.0/npm-left-pad@1.3/
Only in npm-left-pad@1.2.0/npm-left-pad@1.2/package: .npmignore
diff -r npm-left-pad@1.2.0/npm-left-pad@1.2/package/README.md npm-left-pad@1.3.0/npm-left-pad@1.3/package/README.md
32a33,34
> **NOTE:** Characters having code points outside of [BMP plan](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) are considered a two distinct characters. See [#58](https://github.com/stevemao/left-pad/issues/58).
>
35,38d36
<
< ## Typings
<
< Typings copied from [@types/left-pad](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/left-pad) for convenience.
diff -r npm-left-pad@1.2.0/npm-left-pad@1.2/package/index.d.ts npm-left-pad@1.3.0/npm-left-pad@1.3/package/index.d.ts
1c1
< // Type definitions for left-pad 1.1
---
> // Type definitions for left-pad 1.2.0
3,4c3,4
< // Definitions by: Zlatko Andonovski <https://github.com/Goldsmith42>
< // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
---
> // Definitions by: Zlatko Andonovski, Andrew Yang, Chandler Fang and Zac Xu
>
5a6,7
>
> declare namespace leftPad { }
diff -r npm-left-pad@1.2.0/npm-left-pad@1.2/package/package.json npm-left-pad@1.3.0/npm-left-pad@1.3/package/package.json
3c3
<   "version": "1.2.0",
---
>   "version": "1.3.0",
12a13
>     "fast-check": "0.0.8",
diff -r npm-left-pad@1.2.0/npm-left-pad@1.2/package/test.js npm-left-pad@1.3.0/npm-left-pad@1.3/package/test.js
7a8
> var fc = require("fast-check");
51a53,87
> });
>
> var runProperty = function(assert, name, checkFn) {
>   var prop = fc.property(fc.string(), fc.nat(1000), fc.char(), checkFn);
>   var result = fc.check(prop);
>   var message = '';
>   if (result.failed) {
>     message = 'Property "' + name + '" failed on counterexample ' + JSON.stringify(result.counterexample) + ' (seed: ' + result.seed + ')';
>   }
>   assert.strictEqual(message, '', name);
> };
>
> test('properties', function (assert) {
>   assert.plan(4);
>   runProperty(assert, 'starts by ch', function(str, len, ch) {
>       var beg = leftPad(str, len, ch).substr(0, len -str.length);
>       for (var idx = 0 ; idx != beg.length ; ++idx)
>         if (beg[idx] !== ch)
>           return false;
>       return true;
>   });
>   runProperty(assert, 'ends by str', function(str, len, ch) {
>       var out = leftPad(str, len, ch);
>       for (var idx = 0 ; idx != str.length ; ++idx)
>         if (str[str.length -idx -1] !== out[out.length -idx -1])
>           return false;
>       return true;
>   });
>   runProperty(assert, 'len char long if padded (unchanged otherwise)', function(str, len, ch) {
>       var out = leftPad(str, len, ch);
>       return str.length < len ? out.length === len : str === out;
>   });
>   runProperty(assert, 'no ch equivalent to space', function(str, len) {
>       return leftPad(str, len) === leftPad(str, len, ' ');
>   });
```

(It doesn't have to exactly be in diff format, but I suppose that would be helpful so that users could then do stuff with it.)

Currently, the output looks like:

image

Minor, but I think any filenames being output can ignore the absolute directory and just show the relative one.

gfs:
Thanks mike. 

Am a little lost on your comment about context. The example you show doesn’t include any itself. I can add some number of lines of context by default but how many do you want?

It does appear to contain line numbers but I can’t quite parse what exactly those numbers mean in the output you pasted. I can try to emulate that output format though. 


extracting archives is a separate issue but I can make that the default. 

Full paths is a bug. I intended it to be relative paths.

scovetta:
I don't know that line numbers are needed for every line -- the spec for the diff output format is [here](http://kirste.userpage.fu-berlin.de/chemnet/use/info/diff/diff_3.html#SEC13) -- the diff library we're using might provide that, not sure.

One example from above is:
```
diff -r npm-left-pad@1.2.0/npm-left-pad@1.2/package/index.d.ts npm-left-pad@1.3.0/npm-left-pad@1.3/package/index.d.ts
1c1
< // Type definitions for left-pad 1.1
---
> // Type definitions for left-pad 1.2.0
```

So this is showing that when comparing index.d.ts in one version to the other, the only difference is the top line, which changed from `// Type definitions for left-pad 1.1` to `// Type definitions for left-pad 1.2.0`. So context aside, the default output should include that text for that file.

So maybe I'm super confused. :-)

gfs:
The diff library does not provide diff format. It provides line by line.

The current output prints the file name and then a + before any added line or a - before any removed line.

I’ll check out the diff format and match it.

gfs:
Hey Mike,

I see now what the issue you were encountering is. When I refactored to add extraction I mistakenly was reading the names of files as content.  The diff now shows the actual content.

src/oss-diff/DiffTool.cs
jacobmsft: awesome with the sarif and location output, I'll go through those. 
By directory offsetting I mean that the directory was moved. If we have a/b/c moved to d/e/c, how do we figure out if that happened? Also if you are just comparing the file names, a/b/a.txt and c/e/a.txt might be paired up, but not be related at all. 
For recursive extractor, maybe test the file for archive (does recursive extractor have a fast function that tells you if its a valid/well known archive or not?), and extract it if it is. Does not need to be part of the PR, its another feature request. looks like you have already done that part...will go through the changes.

gfs: > By directory offsetting I mean that the directory was moved. If we have a/b/c moved to d/e/c, how do we figure out if that happened?

This is too large to include in this PR.  I'm not sure how important it is.  Refactors that move lots of files around are pretty rare.

> Also if you are just comparing the file names, a/b/a.txt and c/e/a.txt might be paired up, but not be related at all.

I compare the full relative path names.

> For recursive extractor, maybe test the file for archive (does recursive extractor have a fast function that tells you if its a valid/well known archive or not?), and extract it if it is.

This is already done and is how RecursiveExtractor works if you ask it to extract.  If you ask it to extract a non archive it just gives you the content back in a FileEntry object.

scovetta:
LGTM, thanks @gfs!

scovetta (approved):

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,437 @@
...
+
+            [Option('r', "removed-only", Required = false, Default = false,
+                HelpText = "Only show removed lines (and requested context).")]
+            public bool RemovedOnly { get; set; } = false;
+
+            [Option('f', "format", Required = false, Default = "text",
+                HelpText = "Choose output format. (text|sarifv1|sarifv2)")]
+            public string Format { get; set; } = "text";
+
+            [Option('O', "output-location", Required = false, Default = null,
```
jacobmsft, line 71: is this a single file or multiple files? if single file, keep it consistent with:
            [Option('o', "output-file", Required = false, Default = "",

src/oss-diff/DiffTool.cs
```diff
@@ -0,0 +1,437 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using CommandLine;
+using CommandLine.Text;
```
jacobmsft, line 9: capture color and reset if getting sigterm? I Ctrl+C'ed and now my terminal is red :) 
nit though

src/OSSGadget.sln
```diff
@@ -85,6 +87,10 @@ Global
 		{23D17E08-9E95-4448-B75D-BABAF1F3C854}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{23D17E08-9E95-4448-B75D-BABAF1F3C854}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{23D17E08-9E95-4448-B75D-BABAF1F3C854}.Release|Any CPU.Build.0 = Release|Any CPU
+		{17C9F321-8A49-4140-B009-AD476F959D42}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
jacobmsft, line 90: Exception here...

E:\root\OSSGadget\src\oss-diff\bin\Debug\net5.0>.\oss-diff.exe -f sarifv2 -O temp/sarif.sarif pkg:cargo/rand@0.8.3  pkg:cargo/rand@0.8.2
Unhandled exception. System.IO.DirectoryNotFoundException: Could not find a part of the path 'E:\root\OSSGadget\src\oss-diff\bin\Debug\net5.0\temp\sarif.sarif'.
   at System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
   at System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at System.IO.StreamWriter.ValidateArgsAndOpenPath(String path, Boolean append, Encoding encoding, Int32 bufferSize)
   at System.IO.StreamWriter..ctor(String path)
   at System.IO.File.WriteAllText(String path, String contents)
   at Microsoft.CST.OpenSource.DiffTool.DiffTool.RunAsync(Options options) in E:\root\OSSGadget\src\oss-diff\DiffTool.cs:line 428
   at CommandLine.ParserResultExtensions.WithParsedAsync[T](ParserResult`1 result, Func`2 action)
   at Microsoft.CST.OpenSource.DiffTool.DiffTool.Main(String[] args) in E:\root\OSSGadget\src\oss-diff\DiffTool.cs:line 84
   at Microsoft.CST.OpenSource.DiffTool.DiffTool.<Main>(String[] args)

jacobmsft: btw, temp doesnt exist. error is probably because of assuming full path exists

src/oss-diff/DiffTool.cs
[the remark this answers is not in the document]
gfs: Fixed

src/oss-diff/DiffTool.cs
gfs: Fixed

src/OSSGadget.sln
gfs: Fixed.

src/oss-diff/DiffTool.cs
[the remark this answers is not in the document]
gfs: Mike indicated VSM Manager should be threadsafe and it looks okay to me upon basic inspection.

gfs:
I added the ability to pass an archive filename as one of the package url arguments.

PR #210, opened 2021-03-12
Gfs/new rules

gfs:
Improve Sarif Output.


src/oss-characteristics/CharacteristicTool.cs
```diff
@@ -267,7 +267,7 @@ public async Task RunAsync(Options options)
                                 {
                                     { target, analysisResult }
                                 };
-                                var encodedName = Uri.EscapeUriString(Path.GetDirectoryName(target) ?? "unknown");
+                                var encodedName = Uri.EscapeUriString(target ?? "unknown");
```
gfs, line 270: @scovetta Why is this being escaped here?

gfs: This results in the output being all % symbolly.

src/oss-detect-backdoor/oss-detect-backdoor.csproj
```diff
@@ -20,7 +20,7 @@
     <None Remove="Resources\BackdoorRules\backdoor_signatures.json" />
     <None Remove="Resources\BackdoorRules\code_execution.json" />
     <None Remove="Resources\BackdoorRules\crypto_currency.json" />
-    <None Remove="Resources\BackdoorRules\data_exfilatration.json" />
+    <None Remove="Resources\BackdoorRules\data_exfiltration.json" />
```
jacobmsft, line 23: this is not important, but adding rules is frequent, we can publish it and download them if updated from oss-characterstics

src/oss-characteristics/CharacteristicTool.cs
scovetta: I don't think the encoding is needed here.

gfs: Thanks. I’ll remove.

gfs:
@scovetta please double check this didn't miss anything you had in your PR.

src/oss-detect-backdoor/oss-detect-backdoor.csproj
gfs: Something to consider if we end up updating the rules more frequently.

gfs:
@jacobmsft please double check what is being reported in Sarif to determine if we need anything else.

gfs:
I removed the overall file sarif result in favor of individual results for each match.  Do we still need the overall result with just tags and package name?

scovetta:
I think I like the individual output better, this looks good. Only request I have is to change `level` to `warning` for everything coming out of oss-detect-backdoor.

scovetta (approved):

gfs:
> Only request I have is to change level to warning for everything coming out of oss-detect-backdoor.

Done.

scovetta:
I've been running into issues with threading in AppInspector -- my workaround was to set SingleThread = true in the AnalyzeOptions constructor in CharactersticTool.cs. It "fixes" the problem, but I suspect there's a race condition somewhere in ApplicationInspector.

The bug manifests as either a parsing error:

```
// Multithreaded
$.\oss-detect-backdoor.exe sample.tgz
Failed to parse sample.tgz\sample.tar\package\index.js

$.\oss-detect-backdoor.exe sample.tgz
Failed to parse sample.tgz\sample.tar\package\index.js
// OK because there are no rules to find anything in this sample -- index.js is an empty file here.

$
```

Or with different number of findings each time a scan is run.

gfs:
Should be fixed here: https://github.com/microsoft/ApplicationInspector/pull/325

gfs:
That PR fixes an issue with 0 length files. If you are experiencing other issues, lets wait for the new version and if you can repro send me the repro files.

gfs:
I don't see anything obvious in the application inspector code that wouldn't work parallelized (I've done a pass before) so I'll try disabling parallel recursive extractor without disabling the parallelization code in appinspector.

gfs:
@scovetta I double checked recursive extractor and tar and gz are not paralellized

PR #211, opened 2021-03-13
New rules, bump deps, output

scovetta:
This PR adds a few new backdoor rules, bumps dependencies (AppInspector in particular), and improves the output.

### DIFF
--- a/src/oss-characteristics/CharacteristicTool.cs
+++ b/src/oss-characteristics/CharacteristicTool.cs
@@ -14,6 +14,7 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+using Microsoft.CST.RecursiveExtractor;
 using System.Threading.Tasks;
 using static Microsoft.CST.OpenSource.Shared.OutputBuilderFactory;
 using SarifResult = Microsoft.CodeAnalysis.Sarif.Result;
@@ -62,6 +63,14 @@ namespace Microsoft.CST.OpenSource
             [Option('c', "use-cache", Required = false, Default = false,
                 HelpText = "do not download the package if it is already present in the destination directory.")]
             public bool UseCache { get; set; }
+
+            [Option('x', "exclude", Required = false, Default = false,
+                HelpText = "exclude specific files or paths.")]
+            public string FilePathExclusions { get; set; } = "";
+
+            public bool TreatEverythingAsCode { get; set; }
+
+            public bool AllowDupTags { get; set; } = false;
         }
 
         public CharacteristicTool() : base()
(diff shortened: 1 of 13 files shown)


scovetta:
I think we'll have some conflicts here, feel free to trash mine in favor of yours, @gfs.

src/oss-characteristics/CharacteristicTool.cs
```diff
@@ -310,6 +326,42 @@ public async Task RunAsync(Options options)
                                 var encodedName = Uri.EscapeUriString(target ?? "unknown");
                                 var purl = new PackageURL("generic", encodedName);
                                 AppendOutput(outputBuilder, purl, analysisResults);
+                                finalResults.Add(analysisResults);
+                            }
+                        }
+                        else if (File.Exists(target))
```
gfs, line 332: AppInspector can take a file or archive as argument you don’t need to pre extract and then delete.

gfs:
> I think we'll have some conflicts here, feel free to trash mine in favor of yours, @gfs.

I’ll figure out the right way to merge on Monday. Probably will manually pick your changes into mine.

src/oss-characteristics/CharacteristicTool.cs
```diff
@@ -310,6 +326,42 @@ public async Task RunAsync(Options options)
...
+                        {
+                            Extractor e = new Extractor();
+                            var tempDirectory = Path.Join(Path.GetTempPath(), Guid.NewGuid().ToString());
+                            var extractorOptions = new ExtractorOptions()
+                            {
+                                ExtractSelfOnFail = true,
```
gfs, line 338: @scovetta FYI on this behavior. This means if we think it’s an archive and it fails to extract we return the archive itself. That means that it is likely a binary file based on the header. Not sure what AI does with those or if that’s the intention here. If you pass RE a regular non archive file you just get it back in a stream regardless of this option.

scovetta: Got it, thanks for the clarification!

scovetta:
> > I think we'll have some conflicts here, feel free to trash mine in favor of yours, @gfs.
> 
> I’ll figure out the right way to merge on Monday. Probably will manually pick your changes into mine.

Sounds good, I won't make any more structural changes -- maybe a few rule updates, but that's it, for the next couple days.

gfs:
> > > I think we'll have some conflicts here, feel free to trash mine in favor of yours, @gfs.
> > 
> > 
> > I’ll figure out the right way to merge on Monday. Probably will manually pick your changes into mine.
> 
> Sounds good, I won't make any more structural changes -- maybe a few rule updates, but that's it, for the next couple days.

Feel free to make rule updates directly in my PR

gfs:
folded these changes into #210

PR #214, opened 2021-03-23
Add RuntimeIdentifiers to each project.

### LINKED ISSUE #213: Recent OSS Gadget doesn't work on Linux
The latest OSS Gadget Linux builds are erroring out, saying that the framework isn't installed. (It shouldn't need to be-- these should be self-contained builds.)

image

scovetta:
This PR adds RuntimeIdentifiers to each project to enable self-contained cross-project publishing.

From [here](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish):

> --self-contained [true|false]
> 
> Publishes the .NET runtime with your application so the runtime doesn't need to be installed on the target machine. Default is true if a runtime identifier is specified and the project is an executable project (not a library project). For more information, see .NET application publishing and Publish .NET apps with the .NET CLI.
> 

More information in [dotnet/sdk#10902](https://github.com/dotnet/sdk/issues/10902).

This should fix #213.

### DIFF
--- a/Pipelines/core-pipeline.yml
+++ b/Pipelines/core-pipeline.yml
@@ -29,7 +29,7 @@ stages:
   - template: templates/dotnet-test-job.yml
     parameters:
       projectPath: 'src/oss-tests/oss-tests.csproj'
-      arguments: '--runtime linux-x64'
+      arguments: '--runtime win-x64'
 
 - stage: SDL
   dependsOn: Test
--- a/Pipelines/core-pipeline.yml
+++ b/Pipelines/core-pipeline.yml
@@ -29,6 +29,7 @@ stages:
   - template: templates/dotnet-test-job.yml
     parameters:
       projectPath: 'src/oss-tests/oss-tests.csproj'
+      arguments: '--runtime linux-x64'
 
 - stage: SDL
   dependsOn: Test
--- a/Pipelines/core-pipeline.yml
+++ b/Pipelines/core-pipeline.yml
@@ -53,6 +53,7 @@ stages:
       projectName: 'OSSGadget'
       preBuild:
       - template: templates/nbgv-set-version-steps.yml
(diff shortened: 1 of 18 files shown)


gfs (approved):
LGTM but @daalcant should confirm that this will cause the pipeline to publish correctly.

daalcant:
@scovetta @gfs I'm not sure why #213 is broken, but adding a runtime identifier via project properties shouldn't be necessary. We are using `dotnet publish` with a runtime identifier switch already (`-r`) which implies the application will be self-contained. I tested this back when I built the pipelines, but maybe things have changed since then. I'll look into this further.

daalcant:
Documentation just for peace-of-mind: https://docs.microsoft.com/en-us/dotnet/core/deploying/

gfs:
The problem is that the publish command is missing these flags `-p:PublishSingleFile=true --self-contained true` see #217

gfs:
Closed. See #218 instead.

daalcant:
Reopening and converting to draft mode. Will attempt to resolve this and any remaining publishing issues in one PR to reduce the number of times I need to test release artifacts.

daalcant:
Publishing using the PublishSingleFile property is increasing the total unpacked size of a release by 4.5-5x (from ~100 MB to ~500 MB). I believe this is because the .NET runtime is again being bundled with each project built by the solution. I have removed this property for the time being and we can reconsider it in the future (perhaps with a revised project structure).

Leaving this PR in draft mode for a bit longer so I can test the binaries and resolve a couple more issues re: code signing and maintaining Linux file permissions.

daalcant:
Everything should be set here. I won't be able to completely validate the fixes until the sign/release job is run, but prelim tests were a success. Letting checks complete then merging. If any issues arise with the release process then I'll troubleshoot them in the morning.

Fix #213 #215 #216

PR #218, opened 2021-03-24
Fix #217

### LINKED ISSUE #213: Recent OSS Gadget doesn't work on Linux

### LINKED ISSUE #217: OSS Gadget Releases aren't being published as single files.
Lots of files in the linux release.

gfs:
Fix #217 
Fix #213

### DIFF
--- a/Pipelines/core-pipeline.yml
+++ b/Pipelines/core-pipeline.yml
@@ -10,8 +10,7 @@ trigger:
   paths:
     include:
     - src
-    exclude:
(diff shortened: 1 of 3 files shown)


daalcant:
I don't think this will work. Publishing a single file only works with executables, so the Shared library will error out. Will leave additional comments in #213.

Issue #172, opened 2020-11-30
Type initializer error for libgit2sharp on Linux


scovetta:
Here is the error:
```
root@3d51e1d89ad7:/opt/src/work# oss-download -e pkg:github/neuecc/MessagePack-CSharp@v2.2.60
Error downloading GitHub package: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception.
```

Using 0.1.249 [link](https://github.com/microsoft/OSSGadget/releases/download/v0.1.249/OSSGadget_linux_0.1.249.zip) on Ubuntu 20.04.

This seems to be related to https://github.com/libgit2/libgit2sharp/issues/1703.

gfs:
It looks like libgit2sharp isn't supported on 20.04 and there doesn't appear to be urgency to release support.  There are some possible workarounds mentioned in the linked bug and others that I can investigate.

scovetta:
Looks like there is a [fixed version](https://www.nuget.org/packages/LibGit2Sharp/0.27.0-preview-0096) of libgit2sharp out now, but when I tried, I'm now getting:

```
Error downloading GitHub package: failed to mmap. Could not write data: Permission denied failed to mmap. Could not write data: Permission denied|
```

daalcant:
Removed libgit2sharp dep with #228

PR #228, opened 2021-05-30
Backdoors, GitHub Package Manager, cleanup.

scovetta:
This PR has a bunch of updates:
- Add new backdoor detection rules (based on recent attacks seen).
- Add preliminary support for Golang, fix GitHub processors (removed libgit2 dependency) (#172).
- Minor improvements (version sorting, test coverage, nullability)
- Remove self-contained csproj attribute (was preventing building in Visual Studio, debugging, etc.)


src/Shared/OSSGadget.cs
```diff
@@ -117,6 +124,57 @@ protected void SelectOutput(string outputFile)
...
+                var toolName = Path.GetFileNameWithoutExtension(entryAssembly);
+                if (toolName != null)
+                {
+                    return toolName;
+                }
+            }
+            return null;
+        }
+
+        public static async Task<string> GetToolVersion(string? toolName = null)
```
gfs, line 151: Try this

```csharp
public static string GetVersionString()
{            
    return (Assembly.GetExecutingAssembly().GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute), false) as AssemblyInformationalVersionAttribute[])?[0].InformationalVersion ?? "Unknown";        
}
```

src/Shared/Utilities/VersionComparer.cs
```diff
@@ -0,0 +1,79 @@
...
+        public static List<string> Parse(string version)
+        {
+            var parts = new List<string>();
+            var inNumber = false;
+            var curStr = "";
+
+            for (int i = 0; i < version.Length; i++)
+            {
+                var ch = version[i];
+                if (Char.IsNumber(ch) ^ inNumber)
```
gfs, line 24: If you have 12.1.1 doesn't this not work?

- `1` - inNumber = false, Char.IsNumber = true
    - Passes first if, sets curstr to empty.
    - Adds ch to curstr
    - Adds curstr (`1`) to parts
    - Sets inNumber to true
- `2`
    - inNumber = true, Char.IsNumber(ch) is true
    - Skips first if statement
    - adds ch to curstr (now `12`)
    - Passes second if and adds curstr (`12`) to list.  List is now [`1`,`12`].
    - Sets inNumber to true

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -209,74 +210,17 @@ public static async Task<JsonDocument> GetJsonCache(string uri, bool useCache =
...
             }
 
-            // Fallback, leaving it alone
-            if (Logger.IsDebugEnabled)  // expensive string join, avoid unless necessary
-            {
-                Logger.Debug("List is not sortable, returning as-is: {0}", string.Join(", ", versionList));
-            }
-            return versionList;
+            // Split Versions
+            var versionPartsList = versionList.Select(s => VersionComparer.Parse(s)).ToList();
```
gfs, line 221: I don't think you need this.  I tested and the default sort seems to understand version numbers. You should just be able to sort the strings without all this manipulation.

```csharp
> var str1 = "12.1.1";
> var str2 = "2.1.1";
> var list = new List<string>();
> list.Add(str2);
> list.Add(str1);
> list
List<string>(2) { "2.1.1", "12.1.1" }
> list.Sort()
> list
List<string>(2) { "12.1.1", "2.1.1" }
```

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -209,74 +210,17 @@ public static async Task<JsonDocument> GetJsonCache(string uri, bool useCache =
...
+            versionPartsList.Sort(new VersionComparer());
+            return versionPartsList.Select(s => string.Join("", s));
```
gfs, line 223: This joins the versions back together with spaces instead of periods.

src/Shared/PackageManagers/ComposerProjectManager.cs
```diff
@@ -113,11 +118,12 @@ public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL pur
                         versionList.Add(versionObject.Name);
                     }
                 }
-                return SortVersions(versionList.Distinct());
+                return versionList.Distinct();
+                //return SortVersions(versionList.Distinct());
```
gfs, line 122: Left in commented out code.

src/Shared/Utilities/VersionComparer.cs
```diff
@@ -0,0 +1,79 @@
...
+            return parts;
+        }
+
+        /// <summary>
+        /// Compares two lists of strings, using natural numbers where possible.
+        /// </summary>
+        /// <param name="x"></param>
+        /// <param name="y"></param>
+        /// <returns></returns>
+        public int Compare(List<string>? x, List<string>? y)
```
gfs, line 49: I don't think you need this at all.  See other comment about sorting version strings natively.

If you do want to do it this way. I would strongly recommend just converting the numbers into numbers before listing them to make this way simpler.

gfs (changes requested):
Mostly looks good.  Please review the VersionComparer comments.  I believe it will not work as is, and can be replaced with a simple string sort.

src/Shared/Utilities/VersionComparer.cs
gfs: Maybe try this
```csharp
 public static List<string> Parse(string version)
{
    var parts = new List<string>();
    var curStr = string.Empty;

    for (int i = 0; i < version.Length; i++)
    {
        var ch = version[i];
        if (Char.IsNumber(ch))
        {
            curStr += ch;
        }
        else if (ch == '.'){
            parts.Add(curStr);
            curStr = string.Empty;
        }
        else{
            break;
        }
    }
    if (!string.IsNullOrEmpty(curStr)){
        parts.Add(curStr);
    }
    return parts;
}
```

gfs (approved):

src/Shared/PackageManagers/ComposerProjectManager.cs
scovetta: Fixed, thanks!

src/Shared/Utilities/VersionComparer.cs
scovetta:
```
for (int i=0; i<minLength; i++)
{
    if (int.TryParse(x[i], out int xInt) &&
        int.TryParse(y[i], out int yInt))
    {
        if (xInt > yInt) return -1;
        if (yInt > xInt) return 1;
    }
    else
    {
        var compValue = x[i].CompareTo(y[i]);
        if (compValue != 0) return compValue;
    }
}
```

If the strings are integers (e.g. the `234` in `v1.234.56` and the `78` in `v1.78.0`), then it should compare 234 to 78 as integers and show that 234 comes after it.

(Or maybe I missed something.)

src/Shared/PackageManagers/BaseProjectManager.cs
scovetta: There shouldn't be spaces there -- test cases in VersionTests.cs  looks like it should work as expected.

src/Shared/Utilities/VersionComparer.cs
scovetta: I don't this one would work -- we want `a.1.23-rc1` to parse out as `[a, ., 1, ., 23, rc, 1]` so that we can later compare each part separately (e.g.  `a.1.23-rc1` to  `a.1.-rc2`).

scovetta:
@daalcant Could you take a look at the .csproj files (with me setting `<SelfContained>false</SelfContained>`) to see if we need to adjust the build scripts? This was needed to get the solution to build in Visual Studio.

src/Shared/Utilities/VersionComparer.cs
gfs: I tested and it does not sort integers like that when you pass them as strings.  That said the string comparison looks like its only when something is not a number so I think it is OK.  My comment was just explaining why I crossed out my initial comment.
```
> list
List<string>(0) { }
> list.Add("234");
> list.Add("73");
> list.Sort()
> list
List<string>(2) { "234", "73" }
```

daalcant:
> @daalcant Could you take a look at the .csproj files (with me setting `<SelfContained>false</SelfContained>`) to see if we need to adjust the build scripts? This was needed to get the solution to build in Visual Studio.

I'll need to update the build strategy. I think if we build each project individually then do a self-contained build of the solution it will achieve what we're looking for. After I test locally I'll update this PR.

daalcant:
I added some changes to prepare for updating the pipeline build order. I expect builds to fail until I commit the next set of changes; just checking to make sure that tests run as expected.

daalcant:
Running into more issues here. Looks like there has been a recent breaking change to the way self-contained project references are validated in .NET 5. Going to fiddle with the configurations and the `ValidateExecutableReferencesMatchSelfContained` property. 

https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/referencing-executable-generates-error

daalcant:
@scovetta Builds seem to be working again (for now). Unfortunately I'm pretty sure our setup is unsupported and could break in the future. The .NET docs suggest we shouldn't expect to bundle our executables in a single folder the way we have been, which is why we keep running into problems. I'm open to discuss later but I suppose this is fine for the short term.

I performed some basic testing of the framework dependent netcore app and the self-contained Windows and Linux apps. All of the executables run, and I was able to successfully demo `oss-download` against pkg:cargo/rand on each platform as a spot check. Feel free to merge if this is good enough.

scovetta:
Thanks @daalcant! Merging now.

PR #235, opened 2021-06-05
Make commands easier to run in Docker

keeganwitt:
This will allow a user to run a command with something as simple as `oss-download` instead of `./oss-download/bin/Debug/net5.0/oss-download`.

The [wiki page](https://github.com/microsoft/OSSGadget/wiki/Docker-Image) will also need updated to match. I suggest also adding `--rm` to the `run` command. Plus `latest` is the default anyway, so the command could just be ` docker run --rm -it ossgadget` (same as `build`, as far as that goes, i.e. ` docker build -t ossgadget .`).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+/.github/
+/.vs/
+/Pipelines/
+/.gitattributes
+/.gitignore
+/version.json
+/CodeMaid.config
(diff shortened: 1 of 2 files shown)


scovetta (approved):
LGTM, thank you @keeganwitt!

PR #237, opened 2021-06-13
Add oss-reproducible tool

scovetta:
The purpose of the oss-reproducible tool is to determine whether an existing package can be reproduced from its purported source repository. For example, the NPM [left-pad](https://www.npmjs.com/package/left-pad) package is a .tgz file containing some files. The package references a source repo on GitHub. Can we go to the source repo, try to "build" left-pad, and then see that we get the same content?

In this PR, we've implemented a few different strategies:

* Does the source code repo exactly match the package contents?
* Are the package contents contained within the source repo?
* Are the package files and contents contained within the source repo?
* Can we "build" the source repo? (Currently, only NPM and RubyGems have any logic here)
* Can the [Microsoft/Oryx](https://github.com/Microsoft/Oryx) project build the source repo?

We have support for custom build scripts (example is in BuildHelperScripts/npm) and per package and per-strategy ignore filters. We also use Docker containers for the automatic builders, so you'll need to have Docker installed in order to run any of them, or Oryx, which also uses Docker containers.

Example output:
```
> .\oss-reproducible.exe pkg:npm/bluebird@3.7.2

   ____   _____ _____    _____           _            _
  / __ \ / ____/ ____|  / ____|         | |          | |
 | |  | | (___| (___   | |  __  __ _  __| | __ _  ___| |_
 | |  | |\___ \\___ \  | | |_ |/ _` |/ _` |/ _` |/ _ \ __|
 | |__| |____) |___) | | |__| | (_| | (_| | (_| |  __/ |_
  \____/|_____/_____/   \_____|\__,_|\__,_|\__, |\___|\__|
                                            __/ |
                                           |___/
OSS Gadget - oss-reproducible 0.1.288+6d84600a85 - github.com/Microsoft/OSSGadget

Analyzing: pkg:npm/bluebird@3.7.2...
Downloading...
Locating source...
Downloading source...
Out of 5 potential strategies, 4 apply. Analysis will stop after the first successful strategy is found.

Results:
  [✗] PackageMatchesSourceStrategy
  [✗] PackageContainedInSourceByFilenameStrategy
  [✗] PackageContainedInSourceByContentStrategy
  [✓] AutoBuildProducesSamePackage

Summary:
  [✓] Yes, this package is reproducible.
```
We have lots more work to do here.


scovetta:
I've love some feedback on the UI -- Trying to show per-strategy pass/fail, but also the differences (reasons why the strategy fails). The (S) means "this exists in the source repo", (P) "this exists in the package". The +/- is the diff, and the (#)'s are the line number the new line appears in.

I don't love this -- if anyone has has UI/UX as a superpower, I'd be very open to suggestions or a mockup.

image

gfs:
I think

<—
—>

Might be more intuitive to me than P and S

Why are line numbers only on the added lines?

scovetta:
> I think <— —> Might be more intuitive to me than P and S Why are line numbers only on the added lines?

Agree, I added diffoscope, which has more readable text, but has a harder time aligning directories, so we'll need to do some work either way.

Re: The line numbers are only on the added lines -- I think that's because diffplex doesn't include position when ChangeType == Deleted. There's probably a better way of doing it, though.

scovetta:
@gfs Ok no more features for this PR, just a couple days of bug fixes, and I'll plan to merge on Tuesday. We can improve features in the next round of updates.

src/oss-reproducible/Helpers/Helpers.cs
```diff
@@ -0,0 +1,221 @@
...
+                else
+                {
+                    results[hashString] = new HashSet<string>() { filename };
+                }
+            }
+            return results;
+        }
+
+
+        internal static string? GetFirstNonSingularDirectory(string? directory)
```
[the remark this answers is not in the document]
gfs: It appears to:

If the target directory does not exist:
return null
If the target directory has exactly one entry in it and that is a directory
recurse on it
In all other cases
return the directory

src/oss-reproducible/Helpers/Helpers.cs
```diff
@@ -0,0 +1,441 @@
...
+                    return directory;
+                }
+            }
+            else
+            {
+                return directory;
+            }
+        }
+
+        internal static void AddDifferencesToStrategyResult(StrategyResult strategyResult, IEnumerable<DirectoryDifference> directoryDifferences, bool reverseDirection = false)
```
gfs, line 375: IMO this should be a method on StrategyResult itself.

src/oss-reproducible/Helpers/IgnoreFilter.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// <param name="strategyName"></param>
+        /// <param name="filePath"></param>
+        /// <returns></returns>
+        internal static bool IsIgnored(PackageURL? packageUrl, string strategyName, string filePath)
+        {
+            var shouldIgnore = false;
+            filePath = filePath.Replace("\\", "/").Trim();
+            foreach (var filter in FilterText)
+            {
+                var parts = filter.Split(':', 3);
```
gfs, line 64: Specifying 3 here seems unnecessary because you check right after that you have 3 elements.  In fact it would allow you to, for example, silently grab the first 3 elements of an "invalid" 4 element line.

src/oss-reproducible/ReproducibleTool.cs
```diff
@@ -0,0 +1,559 @@
...
+                return KeyValuePair.Create(0.0, "No reproducibility results were created.");
+            }
+
+            var bestScore = KeyValuePair.Create(0.0, "No strategies were able to successfully derive the package from the source code.");
+
+            foreach (var result in fullResult.Results)
+            {
+                var filesString = result.NumIgnoredFiles == 1 ? "file" : "files";
+
+                if (string.Equals(result.StrategyName, "PackageMatchesSourceStrategy"))
```
gfs, line 113: I think strategyname should be an enum instead of a string.

src/oss-reproducible/ReproducibleTool.cs
```diff
@@ -0,0 +1,559 @@
...
+                {
+                    Logger.Warn(ex, "Error processing {0}: {1}", target, ex.Message);
+                    Logger.Debug(ex.StackTrace);
+                }
+            }
+
+            if (finalResults.Any())
+            {
+                // Write the output somewhere
+                //var jsonResults = JsonSerializer.Serialize<List<ReproducibleToolResult>>(finalResults, new JsonSerializerOptions { WriteIndented = true });
```
gfs, line 533: Commented code.

src/oss-reproducible/Strategies/PackageIgnoreList.txt
```diff
@@ -0,0 +1,25 @@
+﻿# All files
```
gfs, line 1: Recommend adding a comment to the top here to explain the format for future users.

src/oss-reproducible/Strategies/PackageMatchesSourceStrategy.cs
```diff
@@ -0,0 +1,65 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.CST.OpenSource.Reproducibility
+{
+    /// <summary>
+    /// This strategy checks to see if the package content exactly matches the source code
+    /// repository. It automatically excludes the .git directory.
```
gfs, line 13: Where is the `.git` directory excluded? its not explicitly excluded by options you set or by the DirectoryDifference method as far as I can tell.

gfs (approved):
Overall looks great!  Left comments where I have feedback.  I think this approach is feasible.

src/oss-reproducible/Helpers/Helpers.cs
```diff
@@ -0,0 +1,221 @@
...
+                throw new DirectoryNotFoundException("Directory does not exist.");
+            }
+
+            var leftFiles = Directory.EnumerateFiles(leftDirectory, "*", new EnumerationOptions { RecurseSubdirectories = true, AttributesToSkip = 0 });
+            // https://stackoverflow.com/questions/2070356/find-common-prefix-of-strings
+            var commonLeftPrefix = new string(leftFiles.First().Substring(0, leftFiles.Min(s => s.Length)).TakeWhile((c, i) => leftFiles.All(s => s[i] == c)).ToArray());
+            leftFiles = leftFiles.Select(s => s[commonLeftPrefix.Length..]);
+
+            var rightFiles = Directory.EnumerateFiles(rightDirectory, "*", new EnumerationOptions { RecurseSubdirectories = true, AttributesToSkip = 0 });
+            var commonRightPrefix = new string(rightFiles.First().Substring(0, rightFiles.Min(s => s.Length)).TakeWhile((c, i) => rightFiles.All(s => s[i] == c)).ToArray());
```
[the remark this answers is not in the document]
scovetta: Fixed in a later commit.

src/oss-reproducible/Strategies/PackageContainedInSourceByContentStrategy.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.CST.OpenSource.Reproducibility
+{
+
+
```
gfs: Whitespace. (Maybe just run code cleanup in Visual Studio?)

src/oss-reproducible/Helpers/Helpers.cs
```diff
@@ -0,0 +1,221 @@
...
+
+            Logger.Debug("GetDirectoryDifferenceByContent({0}, {1})", leftDirectory, rightDirectory);
+            if (leftDirectory == null || rightDirectory == null ||
+                !Directory.Exists(leftDirectory) || !Directory.Exists(rightDirectory))
+            {
+                Logger.Warn("Directory does not exist.");
+                throw new DirectoryNotFoundException("Directory does not exist.");
+            }
+
+            var leftContent = GenerateDirectoryHashes2(packageUrl, strategyName, leftDirectory);
```
gfs: Remove the 2 - it looks like you got rid of the old function and commented it out.

src/oss-reproducible/Helpers/Helpers.cs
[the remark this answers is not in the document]
scovetta: Yep, since RecursiveExtractor tends to create a couple parent directories, we want to ignore them. We basically want to get to the first interesting directory.

src/oss-reproducible/Helpers/IgnoreFilter.cs
```diff
@@ -0,0 +1,92 @@
...
+                var _strategy = parts[1].Trim();
+                var _regex = parts[2].Trim();
+
+                if (string.Equals(_packageManager, "*", StringComparison.InvariantCultureIgnoreCase) ||
+                    string.Equals(_packageManager, packageUrl?.Type ?? "*", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    if (string.Equals(_strategy, "*", StringComparison.InvariantCultureIgnoreCase) || 
+                        string.Equals(strategyName, _strategy, StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        if (Regex.IsMatch(filePath, _regex, RegexOptions.IgnoreCase))
```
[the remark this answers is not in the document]
scovetta: I figured regex can do anything a glob can do and more, no reason to restrict it unless perf is an issue with it, but .NET 5+ should be pretty well optimized for this.

src/oss-reproducible/Strategies/PackageMatchesSourceStrategy.cs
scovetta: Good catch, thanks!

src/oss-reproducible/ReproducibleTool.cs
scovetta: Fixed

src/oss-reproducible/ReproducibleTool.cs
scovetta: Agree, there's some refactoring that this can be a part of -- there's too much strategy-specific refs from ReproducibleTool - let's include this in that work.

src/oss-reproducible/Helpers/IgnoreFilter.cs
scovetta: You should be able to have a 4-element line, e.g.:
```
*:*:^d:/foo/.*
```

(Though I'd expect these to be rare.)

gfs: I don't understand how that is valid.  You only take the first three elements, so you'll be ignoring everything after the 3rd : anyway.

src/oss-detect-backdoor/Resources/BackdoorRules/obfuscation.json
```diff
@@ -84,5 +84,41 @@
...
+    "name": "Backdoor: Long Binary Strings",
+    "id": "BD000805",
+    "description": "Backdoor: Long Binary Strings",
+    "tags": [
+      "Security.Backdoor.Obfuscation.LongBinaryStrings"
+    ],
+    "severity": "moderate",
+    "patterns": [
+      {
+        "pattern": "(\\\\x[a-f0-9][a-f0-9]){10,}",
```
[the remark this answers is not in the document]
scovetta: For this pattern specifically, yes -- we see payloads like "\x12\x34\x56..." - though I suppose we should add an additional pattern for 0xDEADBEEF, etc.

src/oss-reproducible/BuildHelperScripts/npm/autobuild.sh
```diff
@@ -0,0 +1,39 @@
...
+else
+	echo "No custom build script found. Using auto-builder."
+	echo "Executing 'npm install'"
+	npm install
+
+	echo "Executing npm scripts"
+	npm run prepare
+	npm run prepublish
+	npm run publish
+	npm run build
```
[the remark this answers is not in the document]
scovetta: Good catch, fixing in an upcoming commit.

src/oss-reproducible/Helpers/Helpers.cs
```diff
@@ -0,0 +1,441 @@
...
+        /// <param name="leftDirectory">Typically the existing package</param>
+        /// <param name="rightDirectory">Source repo, or built package, etc.</param>
+        /// <returns></returns>
+        public static IEnumerable<DirectoryDifference> DirectoryDifference(string leftDirectory, string rightDirectory, DiffTechnique diffTechnique)
+        {
+            var results = new List<DirectoryDifference>();
+
+            // left = built package, right = source repo
+            var leftFiles = Directory.EnumerateFiles(leftDirectory, "*", SearchOption.AllDirectories);
+            var rightFiles = Directory.EnumerateFiles(rightDirectory, "*", SearchOption.AllDirectories);
```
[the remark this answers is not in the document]
scovetta: Correct, it's a set difference (A - B), since we don't care about additional files in the source repo (right) that aren't present in the package, just the other way around.

src/oss-reproducible/Helpers/IgnoreFilter.cs
scovetta: Sorry, I might be misunderstanding here -- the 2nd param to .Split() is the maximum number of substrings to return, so:

```csharp
foreach (var i in "a:b:c:d:e:f:g".Split(":", 3))
 Console.WriteLine(i);
```
```
a
b
c:d:e:f:g
```

gfs:
I see now. I misunderstood how that parameter version worked.

PR #248, opened 2021-07-01
Fix #245 (output to console on full path)

### LINKED ISSUE #245: Find Source: Output to file fails if we pass a full path
This looks like it affects all tools - we check to see if the path contains any invalid filename chars, but if you pass in a full path, it'll contain slashes, which results in the output going to the console.

scovetta:
This PR has two changes:
- When we accept an output filename, we check to see if it contains invalid chars, and if so, we write to the console instead. (Perhaps we should just error out?). In any case, if the user provides a full path, those slashes cause that check to fail, and the output goes to the screen. This PR fixes that.

- Change banner to stderr (from stdout) - This way, for sarif output (or text for that matter), redirecting can work too.


scovetta:
Thinking about this more, I think we should keep only errors on stderr. We might want a banner suppression flag.

Issue #253, opened 2021-07-06
OSS Characteristic - no matches being returned


scovetta:
We're using tagsOnly, which is resulting in an empty match list. This is breaking oss-detect-backdoor.

PR #254, opened 2021-07-06
OSS Detect Backdoor Improvements

scovetta:
This PR improves some of the backdoor detection rules, as well as #253 which was causing us to come back with empty results. A few other tweaks too.


src/oss-characteristics/CharacteristicTool.cs
```diff
@@ -67,7 +67,7 @@ public static IEnumerable<Example> Examples
                 HelpText = "exclude specific files or paths.")]
```
gfs, line 67:
```suggestion
                HelpText = "exclude files or paths which match provided glob patterns.")]
```

gfs (approved):

Issue #256, opened 2021-07-23
Github projects are downloaded to the wrong location


scovetta:
When oss-download grabs a github project, we add the top level directory twice, so instead of going to /foo/bar/github-x, it goes to /foo/bar/foo/bar/github-x.

PR #257, opened 2021-07-23
Add common crypto library references

scovetta:
This PR improves the oss-detect-cryptography tool in a few days:

First, output is cleaner (run against `pkg:npm/md5.js`):
image

Second, I added a bunch of common crypto libraries in various languages. This means that instead of only detecting implementations of crypto algorithms, we also detect calls to common libraries.

Added a few other bug fixes (like #256).

### DIFF
--- a/src/oss-detect-cryptography/DetectCryptographyTool.cs
+++ b/src/oss-detect-cryptography/DetectCryptographyTool.cs
@@ -523,13 +523,13 @@ namespace Microsoft.CST.OpenSource
 
                     Issue[]? fileResults = null;
                     var task = Task.Run(() => processor.Analyze(buffer, Language.FromFileName(filename)));
-                    if (task.Wait(TimeSpan.FromSeconds(2)))
+                    if (task.Wait(TimeSpan.FromSeconds(30)))
                     {
                         fileResults = task.Result;
                     }
                     else
                     {
-                        Logger.Debug("DevSkim operation timed out.");
+                        Logger.Warn("DevSkim operation timed out.");
                         return analysisResults;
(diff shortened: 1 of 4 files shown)


src/oss-detect-cryptography/DetectCryptographyTool.cs
```diff
@@ -107,55 +108,79 @@ private static async Task Main(string[] args)
...
 
-                            foreach (var t in otherTags)
+                            sb.AppendLine();
+                            sb.AppendLine(Green("Other Cryptographic Characteristics:"));
+                            var characteristics = results.SelectMany(r => r.Issue.Rule.Tags ?? new List<string>())
+                                                         .Distinct()
+                                                         .Where(t => t.Contains("Crypto", StringComparison.InvariantCultureIgnoreCase)&&
+                                                                     !t.StartsWith("Cryptography.Implementation.") &&
+                                                                     !t.StartsWith("Cryptography.Reference."))
+                                                         .Select(t => t.Replace("Cryptography.", ""))
```
gfs, line 171: If I'm not misunderstanding wouldn't this line here get rid of the "Cryptography" portion of the tag that you just found with the Where command.  Why modify the tags?

gfs (approved):

src/oss-detect-cryptography/DetectCryptographyTool.cs
scovetta: Just to make the output cleaner -- a bunch of tags in ApplicationInspector have `Cryptography.` prefixes, which just clutters the screen.

PR #267, opened 2021-11-02
Publish OSS Gadget to Nuget

### LINKED ISSUE #196: Publish OSS Gadget to nuget
We should publish OSS Gadget to nuget as individual libraries for each tool.  We should also distribute each cli tool as a dotnet tool package.

They must be individual to avoid dependency loops.  For example, we want to do something with oss-download in Application Inspector but oss-characteristics uses Application Inspector so we can't include all of oss-gadget.

gfs:
Fix #196.

Creates a template to run and (eventually) a separate pipeline for each ossgadget which will be published. This PR only will publish oss-find-squats but prepares all the csproj files.


daalcant:
*.csproj lgtm

PR #277, opened 2021-11-11
Refactor OSS-FindSquats to publish as Lib

### LINKED ISSUE #273: Remove “pre-release” tag from nuget packages
Nuget packages for oss gadget are being published as prerelease with the commit hash included in version number. This makes is so you have to specify the version to install the tools. 

Ideally this should be changed (though, they do remain "pre-production" quality officially (still 0.1)

### LINKED ISSUE #276: Nuget pipeline is ignoring the paths property
Always runs when a tag is matched.

### LINKED ISSUE #281: Add linting of some kind to the solution.
On the Terrapin team we use [Stylecop](https://github.com/StyleCop/StyleCop) to make sure we follow a consistent code styling and documentation styling in our repository.

I've noticed while poking around the OSSGadget repo, that there's a lot of inconsistencies across the code styling and documentation and so on.

I wouldn't say something like this is "Necessary" or even all that important, but someone who is new to the codebase and trying to understand it and work with it, a consistent styling + linting enforcement would be helpful.

Feel free to shoot me a message on teams or an email about it if you want to see the rules we use in our solution.

gfs:
Refactor Find-Squats into a library.

Enumeration of Squats can now be done by a provided extension method in the find-squats library.  Customization can be done by providing a custom List of mutators.  Npm and Nuget have a slightly different set, with an extensibility point for changing the mutators used by each package manager in the extension method.

To to generate Squats using the default set of mutators for a Project Manager:
```csharp
using Microsoft.CST.OpenSource.FindSquats.ExtensionMethods;
using Microsoft.CST.OpenSource.FindSquats.Mutators;

MutateOptions checkerOptions = new MutateOptions()
{
    SleepDelay = 0
};
var purl = new PackageURL(target);
if (purl.Name is not null && purl.Type is not null)
{
    var manager = ProjectManagerFactory.CreateProjectManager(purl);
    if (manager is not null)
    {
        await foreach (FindSquatResult potentialSquat in manager.EnumerateSquats(purl, checkerOptions))
        {
            ...
        }
    }
}
```

To to generate Squats using a custom set of mutators.
```csharp
using Microsoft.CST.OpenSource.FindSquats.ExtensionMethods;
using Microsoft.CST.OpenSource.FindSquats.Mutators;

MutateOptions checkerOptions = new MutateOptions()
{
    SleepDelay = 0
}; // options are optional to provide
IEnumerable<IMutator> mutators = { your mutators }
var purl = new PackageURL(target);
if (purl.Name is not null && purl.Type is not null)
{
    var manager = ProjectManagerFactory.CreateProjectManager(purl);
    if (manager is not null)
    {
        await foreach (FindSquatResult potentialSquat in manager.EnumerateSquats(purl, mutators, checkerOptions))
        {
            ...
        }
    }
}
```

Now return Mutation objects from the enumeration of Mutations.
Removes manager specific code from FindSquatsTool.
Adds an .editorconfig and runs code cleanup on FindSquats, FindSquatsLib, FindDomainSquats, Tests and Shared.

Also will fix pipeline.

Fix #273
Fix #276
Fix #281


gfs:
@jpinz Is this mechanism acceptable for your use?

gfs:
@cormacpayne

jpinz (changes requested):
Left some comments for ya.

src/oss-find-domain-squats/FindDomainSquatsTool.cs
```diff
@@ -18,12 +18,29 @@
...
+            new DuplicatorMutator(),
+            new PrefixMutator(),
+            new RemovedCharacterMutator(),
+            new SeparatorMutator(),
+            new SubstitutionMutator(),
+            new SuffixMutator(),
+            new SwapOrderOfLettersMutator(),
+            new UnicodeHomoglyphMutator(),
+            new VowelSwapMutator(),
+            new BitFlipMutator()
```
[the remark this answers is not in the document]
gfs: I remember having rearranged this at some point. Maybe that didn't get pushed. Will fix. Agree it should be alpha.

src/oss-find-squats-lib/FindSquatResult.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.CST.OpenSource.FindSquats
+{
+    /// <summary>
+    /// Represents a potential squatted package.
+    /// </summary>
+    public class FindSquatResult
```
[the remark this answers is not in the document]
gfs: Good suggestion.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+            new SwapOrderOfLettersMutator(),
+            new UnicodeHomoglyphMutator(),
+            new VowelSwapMutator()
+        };
+
+        internal static IEnumerable<Mutator> NugetMutators { get; } = BaseMutators.Where(x => x is not UnicodeHomoglyphMutator and not SuffixMutator)
+            .Append(new SuffixMutator(additionalSuffixes: new[] { "net", ".net", "nuget" }, skipSuffixes: new[] { "." }));
+
+        internal static IEnumerable<Mutator> NpmMutators { get; } = BaseMutators.Where(x => x is not UnicodeHomoglyphMutator and not SuffixMutator)
+            .Concat(new Mutator[]
```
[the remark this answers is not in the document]
gfs: It saves the creation of a temporary one element array. You can only do append with one element but concat works with a list.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+            HashSet<string> alreadyChecked = new();
+            if (purl.Name is null || purl.Type is null)
+            {
+                yield break;
+            }
+            foreach(var mutator in mutators)
+            {
+                foreach(var mutation in mutator.Generate(purl.Name))
+                {
+                    if (!alreadyChecked.Add(mutation.Mutated))
```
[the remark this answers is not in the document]
gfs: Sure. That's an easy fix.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+                    {
+                        Logger.Trace($"Already chcked {mutation.Mutated}. Skipping.");
+                        continue;
+                    }
+                    if (options.SleepDelay > 0)
+                    {
+                        Thread.Sleep(options.SleepDelay);
```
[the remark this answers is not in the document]
gfs: Yes. It's just to avoid getting throttled by the package manager you're targeting.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                    var candidatePurl = new PackageURL(purl.Type, mutation.Mutated);
+                    FindSquatResult? res = null;
+                    try
+                    {
+                        var versions = await manager.EnumerateVersions(candidatePurl);
```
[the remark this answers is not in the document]
gfs: I don't know that we can rely on all repositories getting a 404 when a package doesn't exist. For example if we are querying and the result would be json it's possible the response would just an empty valid json object right?

src/oss-find-squats-lib/Mutators/PrefixMutator.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <param name="overridePrefixes">An optional parameter for list of prefixes to replace the default list with.</param>
+        public PrefixMutator(string[]? additionalPrefixes = null, string[]? overridePrefixes = null, string[]? skipPrefixes = null)
+        {
+            if (overridePrefixes != null)
+            {
+                _prefixes = new List<string>(overridePrefixes);
+            }
+            if (additionalPrefixes != null)
+            {
+                _prefixes = new List<string>(_prefixes.Concat(additionalPrefixes));
```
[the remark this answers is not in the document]
gfs: It's an enumerable not a list so it doesn't have Range.

src/oss-find-squats-lib/Mutators/PrefixMutator.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+        }
+
+        public IEnumerable<Mutation> Generate(string arg)
+        {
+            return _prefixes.Select(s => new Mutation(
+                    mutated: string.Concat(s, arg),
+                    original: arg,
+                    mutator: Kind,
+                    reason: "Prefix Added")
```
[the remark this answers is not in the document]
gfs: I can go through and add context to the resin when it makes sense. Like "added suffix: js" instead of just "added suffix"

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                    if (options?.SleepDelay > 0)
+                    {
+                        Thread.Sleep(options.SleepDelay);
+                    }
+                    PackageURL candidatePurl = new(purl.Type, mutation.Mutated);
+                    FindSquatResult? res = null;
+                    try
+                    {
+                        IEnumerable<string>? versions = await manager.EnumerateVersions(candidatePurl);
```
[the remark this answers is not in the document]
gfs: Can you elaborate on why checking here is a problem and what you'd prefer to have?

gfs:
> Left some comments for ya.

Thanks. I expect to have a new version up with those suggestions by this afternoon.

src/oss-find-squats-lib/MutateExtension.cs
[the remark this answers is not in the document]
gfs: I took a quick look at the methods we have implemented so far and I think enumerate versions is still going to be the most robust one.  However, I have added a "PackageExists" method to the BaseProjectManager which (for now) just wraps the EnumerateVersions(purl).Any() check.  The squat checked will use the PackageExists method, which can be optimized in the future.

src/oss-find-squats-lib/MutateExtension.cs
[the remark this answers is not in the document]
gfs: I've refactored this to use a call to a new function: PackageExists.  PackageExists (for now) just uses EnumerateVersions, but can be improved in the future.

gfs:
@jpinz New version with response to all comments.

src/oss-find-squats-lib/MutateExtension.cs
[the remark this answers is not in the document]
gfs: If you can validate that your solution of just a getasync works for all (or many) managers it can override the call for PackageExists for those managers.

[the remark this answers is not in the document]
gfs: Check the latest. I have made a custom implementation for Nuget  and NPM managers that will shortcut enumerating versions. By default a package manager will enumerate versions unless it has a custom implementation for PackageExists.

gfs:
@jpinz I dont think it will make this PR but you mentioned you have your own caching mechanism and want the ProjectManagers to support that. What caching mechanism do you use? Is it a replacement URL? Does it behave the same as the regular managers for those packages?  Discussion for this should continue on #282 .

src/oss-find-squats-lib/MutateExtension.cs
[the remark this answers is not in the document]
gfs: #283 to track implementation for the remaining managers.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -301,6 +363,23 @@ public async Task<string> ExtractArchive(string directoryName, byte[] bytes, boo
...
 
+        /// <summary>
+        /// Check if the package exists in the respository.
+        /// </summary>
+        /// <param name="purl">The PackageURL to check.</param>
+        /// <param name="useCache">Ignored by <see cref="BaseProjectManager"/> but may be respected by inherited implementations.</param>
+        /// <returns>True if the package is confirmed to exist in the repository. False otherwise.</returns>
+        public virtual async Task<bool> PackageExists(PackageURL purl, bool useCache = true)
+        {
+            Logger.Trace("PackageExists {0}", purl?.ToString());
```
jpinz, line 374: Seeing in the log. "PackageExists {purl}" to me, would read as the package exists. However, and I could be wrong, this doesn't seem to be the case? This is just saying we are running `PackageExists` on this package.
This applies to all package managers as far as I can tell.
I would suggest something more like this, unless I am wrong and `Logger.Trace` is a way to say "we are running `PackageExists` on this package"
```suggestion
            Logger.Trace("Checking to see if {0} exists.", purl?.ToString());
```
Or something like that.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -74,6 +75,24 @@ public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl,
...
+        /// <param name="useCache">If cache should be used.</param>
+        /// <returns>True if the package is confirmed to exist in the repository. False otherwise.</returns>
+        public override async Task<bool> PackageExists(PackageURL purl, bool useCache = true)
+        {
+            Logger.Trace("PackageExists {0}", purl?.ToString());
+            if (purl is null || purl.Name is null || purl.Type is null)
+            {
+                Logger.Trace("Provided PackageURL was null.");
+                return false;
+            }
```
jpinz, line 91: It seems that this code is repeated in every `PackageManager`. Could we refactor it somewhere to like a helper method or the method implementation or something? Just to avoid repetition.

src/Shared/PackageManagers/PackageManagerFactory.cs
```diff
@@ -16,7 +16,7 @@ public static BaseProjectManager CreateBaseProjectManager(string destinationDire
         /// </summary>
         /// <param name="purl"> </param>
```
jpinz, line 17: Nit: Docs? Add `destinationDirectory` as well?

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -0,0 +1,117 @@
+﻿using Microsoft.CST.OpenSource.FindSquats.Mutators;
+using Microsoft.CST.OpenSource.Shared;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace Microsoft.CST.OpenSource.FindSquats.ExtensionMethods
+{
+    public static class MutateExtension
```
jpinz, line 10: nit: Docs/comments in this class

src/oss-find-squats-lib/Mutators/AfterSeparatorMutator.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CST.OpenSource.FindSquats.Mutators
+{
+    /// <summary>
+    /// Generates mutations for if a character was changed after a separator was used.
+    /// NOTE: Doesn't currently support extra separators like the <see cref="SeparatorMutator"/> does.
```
jpinz, line 11: I think you added this, so this line is no longer needed.

src/oss-find-squats-lib/Mutators/DoubleHitMutator.cs
```diff
@@ -0,0 +1,37 @@
...
+                    yield return new Mutation(
+                        mutated: string.Concat(arg[..i], c, arg[i..]),
+                        original: arg,
+                        mutator: Kind,
+                        reason: $"Double hit character: {c}");
+                    yield return new Mutation(
+                        mutated: string.Concat(arg[..(i + 1)], c, arg[(i + 1)..]),
+                        original: arg,
+                        mutator: Kind,
+                        reason: $"Double hit character: {c}");
```
jpinz, line 32: Looking over this now, I'm trying to figure out what the difference is?
Isn't this just adding the doubled character before, and then another time, after? I think that would be the same thing right?
Like loda **s** sh, and lodas **s** h? It's the same string but the s was duplicated once before the s, and another mutation with the duplicated s after the s.

src/oss-find-squats-lib/Mutators/RemovedCharacterMutator.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public IEnumerable<Mutation> Generate(string arg)
+        {
+            for (int i = 1; i < arg.Length; i++)
+            {
+                yield return new Mutation(
+                    mutated: $"{arg[..i]}{arg[(i + 1)..]}",
+                    original: arg,
+                    mutator: Kind,
+                    reason: $"Character Removed: {i}");
```
jpinz, line 22:
```suggestion
                    reason: $"Character Removed: {arg[i]}");
```

src/oss-find-squats-lib/Mutators/SuffixMutator.cs
```diff
@@ -0,0 +1,48 @@
...
+                _suffixes.RemoveAll(x => skipSuffixes.Contains(x));
+            }
+        }
+        public IEnumerable<Mutation> Generate(string arg)
+        {
+            return _suffixes.Select(s => new Mutation(
+                    mutated: string.Concat(arg, s),
+                    original: arg,
+                    mutator: Kind,
+                    reason: $"Suffix Added {s}"));
```
jpinz, line 45:
```suggestion
                    reason: $"Suffix Added: {s}"));
```

jpinz (changes requested):
@gfs looks much better! I left a handful more comments.
I think after this we should be good to go, or at the very least good to give it a shot and see what other things come up once I try to use it in our code base,

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -1,27 +1,27 @@
...
+    using System.Net.Http;
+    using System.Text;
+    using System.Text.Json;
+    using System.Text.RegularExpressions;
+    using System.Threading.Tasks;
+    using F23.StringSimilarity;
+    using Microsoft.CST.OpenSource.Model;
+    using Microsoft.CST.RecursiveExtractor;
+    using Microsoft.Extensions.Caching.Memory;
+    using Version = SemanticVersioning.Version;
```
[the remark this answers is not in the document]
gfs: This was a styling correction. There are edge cases where it is relevant. 

https://stackoverflow.com/questions/125319/should-using-directives-be-inside-or-outside-the-namespace

src/Shared/PackageManagers/BaseProjectManager.cs
gfs: The trace messaging is consistent with other trace calls to simply indicate entrance to the method.

End users are not expected to ever see the trace calls - this should only ever be if you are debugging the behavior and need to see which calls are being made.

src/Shared/PackageManagers/NPMProjectManager.cs
gfs: Unfortunately I think if we refactor this out we lose the ability to trace the nullability and later to satisfy the compiler would need to handle purl remaining nullable (even though the helper method should confirm it is not nullable.) 

Perhaps it could be worked around with a method like:

bool PurlIsValid(PackageURL purl, out PackageURL notNullPurl)

But you can also note that it's not *exactly* the same for each because some managers care about, for example; the namespace field, and some do not.

src/oss-find-squats-lib/Mutators/DoubleHitMutator.cs
gfs: I believe I copied this from your PR and also am unclear on what the second one does. Can remove.

src/oss-find-squats-lib/Mutators/RemovedCharacterMutator.cs
gfs: I actually think the index of the removed character is clearer than the content of the removed character in this case. But I could potentially include both.  

The Mutation object will contain both the original and the mutated version.

gfs:
I'll wrap up these comments and I'm also working on a code styling commit for the Shared lib and then I'll merge this hopefully this evening and it should be published to nuget.

src/Shared/PackageManagers/BaseProjectManager.cs
gfs: Perhaps it would be clearer to change this to say 

`Logger.Trace("PackageExists(args1,arg2")`

With the appropriate arguments.

src/Shared/PackageManagers/CPANProjectManager.cs
```diff
@@ -22,6 +22,24 @@ public CPANProjectManager(string destinationDirectory) : base(destinationDirecto
...
+        /// <summary>
+        /// Check if the package exists in the respository.
+        /// </summary>
+        /// <param name="purl">The PackageURL to check.</param>
+        /// <param name="useCache">If cache should be used.</param>
+        /// <returns>True if the package is confirmed to exist in the repository. False otherwise.</returns>
+        public override async Task<bool> PackageExists(PackageURL purl, bool useCache = true)
+        {
+            Logger.Trace("PackageExists {0}", purl?.ToString());
+            if (purl is null || purl.Name is null || purl.Type is null)
```
cormacpayne, line 34: nit: do we need to check if `purl.Name` and/or `purl.Type` are also empty strings? Also do we need to check if `purl.Type` is `null` if we're not using it in this function?

(Looks like this comment can be applied to most project manager classes, so feel free to do what you'd like with this comment 😀)

src/oss-find-squats-lib/FindPackageSquatResult.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Microsoft.CST.OpenSource.FindSquats.Mutators;
```
cormacpayne, line 1: Looks like the copyright header is missing in a couple of new files (I _think_ with the new linting rules you have in place, you can enforce a copyright header)

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: Thanks for the info! That's what I figured, just wasn't sure.

src/oss-find-squats-lib/Mutators/DoubleHitMutator.cs
jpinz: I copied it from the original `Generative.cs` file 😅

src/oss-find-squats-lib/Mutators/AfterSeparatorMutator.cs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CST.OpenSource.FindSquats.Mutators
+{
+    /// <summary>
+    /// Generates mutations for if a character was changed after a separator was used.
```
cormacpayne, line 10: nit: I believe I left it as a generic comment on @jpinz's PR, but in the future it would be nice to have samples of the mutator in its class doc comment

src/oss-find-squats-lib/Mutators/AfterSeparatorMutator.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+
+        public IEnumerable<Mutation> Generate(string arg)
+        {
+            foreach (char s in Separators)
+            {
+                string[]? splits = arg.Split(s, StringSplitOptions.RemoveEmptyEntries);
+
+                if (splits.Length == 2)
```
cormacpayne, line 41: nit: this mutator is only checking packages with a _single_ instance of a separator in its name, correct? For example, a package like `Newtonsoft.Json.Schema` would never return any mutations here, but `Newtonsoft.Json` may return some.

cormacpayne (approved):
:shipit:

src/oss-find-squats-lib/Mutators/DoubleHitMutator.cs
gfs: I see that this is the case in main as well, but I still do not remember why we did this.

src/Shared/PackageManagers/CPANProjectManager.cs
gfs: I think I was overzealous in checking Type and that can in fact be removed.

If it is an empty string the later call will likely fail. So it probably does make sense to check that here as well as the purpose of this check is to shortcircuit actually making a HTTP call for something we know won't return any results.

src/oss-find-squats-lib/Mutators/AfterSeparatorMutator.cs
gfs: Fixed.

src/oss-find-squats-lib/FindPackageSquatResult.cs
gfs: Found a way to do with editorconfig. Thanks!

src/oss-find-squats-lib/Mutators/AfterSeparatorMutator.cs
jpinz: Definitely worth doing a future doc/examples pass on all the mutators.

gfs: https://github.com/microsoft/OSSGadget/issues/284

gfs:
Packages are now published and seem to be working.

https://www.nuget.org/packages/Microsoft.CST.OSSGadget.FindSquats/

PR #280, opened 2021-11-16
Changes to how Mutators are used.

jpinz:
_Summarize the changes made and why they are needed._
 
I created this PR to simplify how mutators work and how they are implemented.
I did this by moving the mutators all to the shared library in models, and adding a list of mutators to each package manager to specify which mutators are implemented for that package manager.
For example, NuGet and NPM are both not using the `UnicodeHomoglyphMutator.cs` mutator.

I did cherry-pick some of the files from #277, the `oss-find-squats-lib` files for SquatChecker, and FindSquatsResult, as well as the changes for FindSquatsTool in `oss-find-squats`.

## Testing
_Summarize what testing (automated or manual) has been done to verify this change._
 
## Checklist
_Please complete the following before publishing this PR. For any below items that are not applicable or otherwise can’t be completed, 
strike out item (place `~~` around it on both sides) and add a note to the end explaining why this can’t be done._

- [ ] Unit Testing: Changed code is covered by passing unit tests.
- [ ] Acceptance Testing: Changes to API definitions and/or behavior are covered by passing acceptance / integration tests.
- [X] Code Comments: Where not self-evident, changed code includes comments describing its purpose and/or why it was written a certain way.
- [ ] Logging & Metrics: This change has sufficient logging & metrics for us to diagnose problems in production.

## TODO:

- [ ] Add BitFlip mutator back.
- [ ] Maybe more mutators?


gfs:
@jpinz Thanks for the contribution. I'll give this a close look tomorrow.

I do like giving the mutators each their own class, but I'm less sure about adding mutator information to each package manager - that seems to be to be an easy way to get unexpected behavior if you don't remember to add a new mutator to one to the N package managers.

jpinz:
> @jpinz Thanks for the contribution. I'll give this a close look tomorrow.
> 
> I do like giving the mutators each their own class, but I'm less sure about adding mutator information to each package manager - that seems to be to be an easy way to get unexpected behavior if you don't remember to add a new mutator to one to the N package managers.

Thanks for checking it out!
I get the concern, I have a feeling we can probably find a decent solution to that. Also, a good argument for detailed unit tests 😄 
The PR is in draft because I anticipated a lot of good conversation stemming from the initial publish of just this draft. So I am wide open to suggestions from everyone!

src/Shared/Model/Mutators/BaseMutator.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.CST.OpenSource.Model.Mutators
+{
+    /// <summary>
+    /// The base mutator to be implemented by other mutators.
+    /// </summary>
+    public abstract class BaseMutator
+    {
+        public string Mutator = "BASE_MUTATOR";
```
cormacpayne, line 15: Should we use an `Enum` to represent these mutator values?

src/Shared/Model/Mutators/BaseMutator.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        /// <summary>
+        /// Generates the typo squat mutations for a string.
+        /// </summary>
+        /// <param name="arg">The string to generate mutations for.</param>
+        /// <returns>A list of mutations with the name and reason.</returns>
+        public abstract IEnumerable<(string Name, string Reason)> Generate(string arg);
```
cormacpayne, line 22: What do you think about representing the output of this function as a custom `Mutation` object that contains this information (and possibly more in the future)? It should make working with this content a little bit easier and more concise with additional documentation in the class.

src/Shared/Model/Mutators/SeparatorMutator.cs
```diff
@@ -0,0 +1,58 @@
...
+        public SeparatorMutator(char[]? additionalSeparators = null, char[]? overrideSeparators = null)
+        {
+            if (overrideSeparators != null)
+            {
+                separators = new HashSet<char>(overrideSeparators);
+            }
+            if (additionalSeparators != null)
+            {
+                separators = new HashSet<char>(separators.Concat(additionalSeparators));
+            }
```
cormacpayne, line 37: I personally think it makes sense to remove the `additionalSeparators` functionality from the constructor and have a separate method that allows the user to provide more separators if they'd like. It's a little confusing have two separate flows intertwined within the same constructor.

src/Shared/Model/Mutators/SeparatorMutator.cs
```diff
@@ -0,0 +1,58 @@
...
+// --------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CST.OpenSource.Model.Mutators
+{
+    /// <summary>
+    /// Generates mutations of added or removed separators.
+    /// Separators are '.', '-', '_'.
```
cormacpayne, line 12: Maybe clarify here and say that these are the default separators, but can be modified.

src/Shared/Model/Mutators/AfterSeparatorMutator.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public override IEnumerable<(string Name, string Reason)> Generate(string arg)
+        {
+            foreach (var s in SeparatorMutator.separators)
+            {
+                var splits = arg.Split(s, StringSplitOptions.RemoveEmptyEntries);
+
+                if (splits.Length == 2)
+                {
+                    for (var c = 'a'; c <= 'z'; c++)
```
cormacpayne, line 26: Any reason we don't want to also iterate over uppercase letters or numbers?

src/Shared/Model/Mutators/PrefixMutator.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        private static IList<string> _prefixes = new List<string>(){ ".", "x", "-", "X", "_" };
+
+        /// <summary>
+        /// Initializes a <see cref="PrefixMutator"/> instance.
+        /// Optionally takes in a additional prefixes, or a list of overriding prefixes to replace the default list with.
+        /// </summary>
+        /// <param name="additionalPrefixes">An optional parameter for extra prefixes.</param>
+        /// <param name="overridePrefixes">An optional parameter for list of prefixes to replace the default list with.</param>
+        public PrefixMutator(string[]? additionalPrefixes = null, string[]? overridePrefixes = null)
```
cormacpayne, line 27: Same comment made previously about dropping `additionalPrefixes` in favor of a separate function.

src/Shared/Model/Mutators/PrefixMutator.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CST.OpenSource.Model.Mutators
+{
+    /// <summary>
+    /// Generates mutations for if a prefix was added to the string.
+    /// We check for these prefixes: ".", "x", "-", "X", "_".
```
cormacpayne, line 13: Any reason these are the default prefixes?

src/Shared/Model/Mutators/RemovedCharacterMutator.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// <summary>
+    /// Generates mutations for removing a character in the string.
+    /// </summary>
+    public class RemovedCharacterMutator : BaseMutator
+    {
+        public new string Mutator = "REMOVED_CHARACTER";
+
+        public override IEnumerable<(string Name, string Reason)> Generate(string arg)
+        {
+            for(var i = 1; i < arg.Length; i++)
```
cormacpayne, line 18: We don't try and remove the first character?

gfs:
I'm still working on the best way to integrate changing the mutators to be classes instead of methods.  I think my main concern with bringing the mutators into the shared library is
1. They're not really shared
2. The way the pipelines work any changes to shared will trigger a rebuild of all of the tools. For now we are only publishing find-squats but eventually that might not be the case. Given that those other components do not actually use the mutators it would create a lot of churn anytime the mutators were updated.

I think I will be keeping the mutators in the FindSquatsLib, but its possible maybe to implement it as an extension method that is available for the packagemangers in the find squats lib.

src/Shared/Model/Mutators/SuffixMutator.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        private static IList<string> _suffixes = new List<string>(){ "s", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "ng", "-ng", ".", "x", "-", "_"};
+
+        /// <summary>
+        /// Initializes a <see cref="SuffixMutator"/> instance.
+        /// Optionally takes in a additional suffixes, or a list of overriding suffixes to replace the default list with.
+        /// </summary>
+        /// <param name="additionalSuffixes">An optional parameter for extra suffixes.</param>
+        /// <param name="overrideSuffixes">An optional parameter for list of suffixes to replace the default list with.</param>
+        public SuffixMutator(string[]? additionalSuffixes = null, string[]? overrideSuffixes = null)
```
cormacpayne, line 27: Same comment about `additionalSuffixes` here

src/Shared/Model/Mutators/SuffixMutator.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CST.OpenSource.Model.Mutators
+{
+    /// <summary>
+    /// Generates mutations for if a suffix was added to the string.
+    /// We check for these prefixes: ".", "x", "-", "X", "_".
```
cormacpayne, line 13: Same comment about how we decided on these prefixes

src/Shared/Model/Mutators/SwapOrderOfLettersMutator.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    public class SwapOrderOfLettersMutator : BaseMutator
+    {
+        public new string Mutator = "LETTERS_SWAPPED";
+
+        public override IEnumerable<(string Name, string Reason)> Generate(string arg)
+        {
+            if (arg.Length < 5)
+            {
+                // TODO: Figure out why we only check for less than 5 char strings.
```
cormacpayne, line 20: I'd be interested in learning what is restricting us to five or more character names

src/oss-find-squats-lib/FindSquatResult.cs
```diff
@@ -0,0 +1,24 @@
+// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// --------------------------------------------------------------------------------------------
+
+using Microsoft.CST.OpenSource.Shared;
+using System.Collections.Generic;
+
+namespace Microsoft.CST.OpenSource.FindSquats
+{
+    public class FindSquatResult
```
cormacpayne, line 10: Docs for this class?

src/oss-find-squats-lib/SquatChecker.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// </summary>
+        /// <returns>A dictionary of each mutation, with a list of the reasons why this mutation showed up.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if the package doesn't have a name.</exception>
+        public Dictionary<string, IList<string>> Mutate(PackageURL pUrl, BaseProjectManager manager)
+        {
+            Dictionary<string, IList<string>> mutations = new();
+
+            // Go through each mutator in this package manager and generate the mutations.
+            var mutationsList = manager.Mutators.SelectMany(m =>
+                m.Generate(pUrl.Name ?? throw new InvalidOperationException()));
```
cormacpayne, line 42: Can we do the null check outside of this `Generate` call? In my opinion it's a little cleaner to handle it at the beginning of the function and then continue with the confirmation that the name is valid.

src/oss-find-squats-lib/SquatChecker.cs
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+
+            // Return the dictionary of all mutations.
+            return mutations;
+        }
+
+        public async IAsyncEnumerable<FindSquatResult> CheckSquats(PackageURL purl, Options? options)
+        {
+            if (purl.Name is not null && purl.Type is not null)
```
cormacpayne, line 63: Another minor style thing I would suggest:

```cs
if (purl?.Name == null || purl.Type == null)
{
    return;
}
```

Returning early will allow us to confirm that `purl` and its `Name` and `Type` properties aren't `null` and then the rest of the code in this function doesn't need to be contained within this if-statement.

src/oss-find-squats-lib/SquatChecker.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        public static NLog.ILogger Logger { get; set; } = NLog.LogManager.GetCurrentClassLogger();
+
+
+        /// <summary>
+        /// Generate the mutations for the package.
+        /// </summary>
+        /// <returns>A dictionary of each mutation, with a list of the reasons why this mutation showed up.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if the package doesn't have a name.</exception>
+        public Dictionary<string, IList<string>> Mutate(PackageURL pUrl, BaseProjectManager manager)
```
cormacpayne, line 36: Does it make sense to move this functionality into the respective `BaseProjectManager` classes (or even the shared logic into the `BaseProjectManager` class itself)? That way the project managers are responsible for implementing (or reusing) the logic for mutating the package names and searching their public registry for the possible packages. That way you don't have to have project manager-specific code in this generic file (_i.e._, we don't have to check if the manager is NuGet to perform an additional check, like we're doing below)

cormacpayne (changes requested):
Overall this looks great -- one thing that would really help is adding an example or two of the mutation that you're trying to capture in each of the `BaseMutator` implementation's docs. I found myself having to piece together what was actually going on because the name of the mutator wasn't enough to describe what the mutation was, so an example or two with some package names would really help.

src/Shared/Model/Mutators/BaseMutator.cs
gfs: I like this idea.

src/Shared/Model/Mutators/AfterSeparatorMutator.cs
gfs: Most (all?) package managers don't care about casing.

Numbers is a good idea.

src/Shared/Model/Mutators/SwapOrderOfLettersMutator.cs
gfs: Combinatorial complexity is my guess. By default this hits the real production servers to query packages.

src/oss-find-squats-lib/SquatChecker.cs
gfs: See my earlier comment on this. I'm uncomfortable putting the mutator functionality into the shared library because it is not shared by any other caller to the package managers other than find squats and changing shared implicitly changes all ~15 tools.

Further the mutators themselves may be used to mutate other things (for example we have a domain squat checker, and it's possible we could find other applications) so tightly coupling them to the package managers doesn't work in that aspect either.

My current thinking is to extend  the class based mutators as extension methods on BasePackageManager in the new find squats lib. 

For filtering, potentially the mutator itself could be responsible then for ensuring it's only run on the package managers that are applicable. Accomplishing the same goal of removing manager specific code from the squat checker itself. 

That said I don't totally disagree with keeping the manager specific complexity inside the squat checker as that is the only time it is relevant.

src/Shared/Helpers/QwertyKeyboardHelper.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        private static string[] _keymap = {"1234567890-=", "qwertyuiop[]\\", "asdfghjkl;'", "zxcvbnm,./",};
+
+        private static Dictionary<char, int> _locations = new()
+        {
+            ['1'] = 0,
+            ['2'] = 1,
+            ['3'] = 2,
+            ['4'] = 3,
+            ['5'] = 4,
```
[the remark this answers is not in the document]
gfs: Since it works as is, reimplementing this is low priority for me. But if you guys wanna take a crack at it I don't mind this proposal.

gfs:
I incorporated a lot of this into #277 with the extension method implementation.

src/Shared/Model/Mutators/SwapOrderOfLettersMutator.cs
gfs: Took a closer look at this function and its a O(N) loop, I dont see any reason to keep the <5.

gfs:
More updates to #277 including refactoring to use a `MutatorType` enum and returning `Mutation` objects.

jpinz:
Closing as #277 included most of this work.

PR #299, opened 2022-01-24
Include any pipeline change.

gfs:


jpinz:
:shipit:

gfs:
It looks like I made the logic for "should we publish this to nuget" a bit too complex and the first attempt to fix it wasn't enough. This simplifies it so it will publish a new version on any pipeline or source change.

jpinz:
Makes sense to me!

PR #300, opened 2022-01-25
Added toJson to FindPackageSquatResult

jpinz:
For our project, we need to be able to store the squat results in cosmos, which we are doing as a string. 
Having a `toJson` method in the `FindPackageSquatResult` would be helpful. We could do it just on our side, but I thought it made sense to have in the OSSGadget project as well.
Could be `toString` instead of `toJson` if you think that's appropriate, the general premise is the same.
Also now includes a `FromJsonString` to construct a `FindPackageSquatResult` from a json string.


gfs:
I fixed the pipeline issue for forks. This seems fine.

jpinz:
@gfs I added a FromJsonString as well, otherwise I think this should be good.

gfs:
Trigger pipelines.

PR #301, opened 2022-01-31
Update Microsoft.Extensions.Caching.Memory version to 6.0.0

jpinz:
Change the version of Microsoft.Extensions.Caching.Memory to the non-preview version of 6.0.0


jpinz:
@gfs we just try to avoid prerelease dependencies where we can.

PR #302, opened 2022-02-01
Added mutation support to the scope for scoped npm packages.

jpinz:
For an npm package e.g. `@angular/core` the typo squats were generating mutations that were just variations on `core` returning things like `coree` and `cort`, etc. but for scoped npm packages we want the mutations to be more like `@engular/core` and `@angullar/core`. With the mutations on the scope/namespace, not the package name.


gfs:
LGTM.

PR #303, opened 2022-02-24
Update Metadata Normalization

jpinz:
This PR aims to add some new features to the metadata tool.
It adds support for NuGet packages.
It adds a homepage property.
Fixes some bugs with NPM package parsing.
Adds a unique value for the frontend package and package version urls and the api urls. ie https://www.nuget.org/packages/Newtonsoft.Json vs https://api.nuget.org/v3/registration5-gz-semver2/newtonsoft.json/index.json

### DIFF
--- a/src/Shared.CLI/Shared.CLI.csproj
+++ b/src/Shared.CLI/Shared.CLI.csproj
@@ -49,7 +49,7 @@
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="6.0.0" />
     <PackageReference Include="NLog" Version="4.7.13" />
     <PackageReference Include="NLog.Schema" Version="4.7.13" />
-    <PackageReference Include="NuGet.Versioning" Version="6.0.0" />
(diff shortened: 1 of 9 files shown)


gfs:
I'll review and merge on Monday if I don't see any issues.

gfs:
I don't think ossmetadata is being published to nuget yet. Would you like a nuget releases like find squats? Would you want lib or cli?

jpinz:
> I don't think ossmetadata is being published to nuget yet. Would you like a nuget releases like find squats? Would you want lib or cli?

@gfs we don't need the oss-metadata tool, all of the metadata logic is already in the Shared lib which is on NuGet so I think we are all set.
I already have a poc working in a branch on our repo using the metadata logic. We just needed NuGet support and I found+fixed a few bugs along the way. As well as adding some additional properties to metadata that we will need on our side at some point. 
So assuming this looks good to you, it should be good to merge and just publish a new version of the Shared lib on NuGet.

gfs:
Okay sounds good. That’s easy enough then. I’ll try to review and merge this today.

PR #304, opened 2022-02-28
Switch to https://www.nuget.org/packages/packageurl-dotnet/

jpinz:
The official package-url dotnet implementation just got an update to have it pushed to nuget, switched from the copied implementation to the one on NuGet.


jpinz:
> Commenter does not have sufficient privileges for PR 304 in repo microsoft/OSSGadget

@gfs could you run it for this PR too? We just switched to the nuget package on our side too.

gfs:
Will fix the missing import from the merge this evening.

jpinz:
The PackageUrl dotnet implementation doesn't have support for Json deserialization like the implementation that was pasted and slightly modified in OSSGadget, so I just put a PR out on the repo to add support for it: https://github.com/package-url/packageurl-dotnet/pull/16

gfs:
> The PackageUrl dotnet implementation doesn't have support for Json deserialization like the implementation that was pasted and slightly modified in OSSGadget, so I just put a PR out on the repo to add support for it: [package-url/packageurl-dotnet#16](https://github.com/package-url/packageurl-dotnet/pull/16)

If they dont approve that PR (they may not because it uses newtonsoft instead of system.text.json) it can just be added back as Extension methods in ossgadget.

jpinz:
> > The PackageUrl dotnet implementation doesn't have support for Json deserialization like the implementation that was pasted and slightly modified in OSSGadget, so I just put a PR out on the repo to add support for it: [package-url/packageurl-dotnet#16](https://github.com/package-url/packageurl-dotnet/pull/16)
> 
> If they dont approve that PR (they may not because it uses newtonsoft instead of system.text.json) it can just be added back as Extension methods in ossgadget.

Updating it to system.text.json now.

gfs:
I'm seeing 3 download tests failing after your last change (despite rerunning the failed tests). 

I'll double check the failures locally tonight.

gfs:
I'm not sure this is related to the changes in the PR, but it seems like some of the download tests are failing.

The cocoa pods test finds a pod using an md5 of the package name and then looks up the url here: https://github.com/CocoaPods/Specs. However, for the test case the MD5 comes out with `8/d/6` but the actual path on the repo is `8/2/7`: https://github.com/CocoaPods/Specs/blob/master/Specs/8/2/7/RandomKit/5.2.1/RandomKit.podspec.json.

https://github.com/microsoft/OSSGadget/blob/a0f6af11f8648f1fd363f293976778048cf2bf75/src/Shared/PackageManagers/CocoapodsProjectManager.cs#L174-L192

I haven't investigated the CRAN or CPAN failures yet.

scovetta:
Nope, while the tests we have definitely could be better, the problem here was that the [PackageUrl](https://github.com/microsoft/OSSGadget/blob/main/src/Shared/PackageUrl/PackageUrl.cs#L309) class that we had embedded included some customizations, e.g.:

image

The problem is that PackageURL toLower()'s the name, but when you toLower() RandomKit, you get randomkit, which MD5's to 8/d/6, which is wrong. So for a couple package types (cpan, cran, cocoapods, and vsm) we need to leave the names in whatever case they came in as.

image

Similarly below in that same file for namespaces.

IIRC, that was the reason I didn't use the existing package.

gfs:
> Nope, while the tests we have definitely could be better, the problem here was that the [PackageUrl](https://github.com/microsoft/OSSGadget/blob/main/src/Shared/PackageUrl/PackageUrl.cs#L309) class that we had embedded included some customizations, e.g.:
> 
> image
> 
> The problem is that PackageURL toLower()'s the name, but when you toLower() RandomKit, you get randomkit, which MD5's to 8/d/6, which is wrong. So for a couple package types (cpan, cran, cocoapods, and vsm) we need to leave the names in whatever case they came in as.
> 
> image
> 
> Similarly below in that same file for namespaces.
> 
> IIRC, that was the reason I didn't use the existing package.

Was this the only change required? we could open a PR upstream to see if they would take this change.

gfs:
https://github.com/package-url/packageurl-dotnet/pull/18

scovetta:
Yep, makes sense to try to upstream. Though IIRC some of the types aren't in the spec (cpan is unofficial, and I just made up vsm for Visual Studio Marketplace). Not that they aren't reasonable to include / add to spec.

gfs:
I opened a PR to add those types to the spec as well.  It looks like there are a number of implementations - not just the C# one, so I'm not sure if those would also require this change to the case sensitivity handling but I think that's out of scope for us. https://github.com/package-url/purl-spec/pull/155/files

src/Shared/Extensions/PackageUrlExtension.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.CST.OpenSource.Extensions;
+
+using System.IO;
+using System.Text.RegularExpressions;
+using PackageUrl;
+
+public static class PackageUrlExtension
+{
+    public static string ToStringFilename(this PackageURL packageUrl)
```
bhuwan-agarwal, line 11: Is there a unit test for this method?

gfs:
> [package-url/packageurl-dotnet#18](https://github.com/package-url/packageurl-dotnet/pull/18)

This has been merged. I assume they will publish a new nuget soon @jpinz.

src/Shared/Extensions/PackageUrlExtension.cs
jpinz: Just added 2 tests.

jpinz:
@gfs just pinged the maintainer of the packageurl-dotnet repo asking if they can publish a new version when they get a chance, so we can unblock this PR.

gfs:
LGTM. 

@jpinz is this ready to merge?

jpinz:
> LGTM.
> 
> @jpinz is this ready to merge?
Should be good! Hoping to get the other PR in today/tomorrow as well and consuming the new version that contains both changes.

PR #305, opened 2022-03-08
Fixed bugs with NPM metadata parsing

jpinz:
Packages without the description on the package level, only on the package version level weren't able to be parsed.
Scoped packages weren't getting the correct metadata, for example `@elliemae/ds-modal` would get the metadata for `ds-modal`

Also switched the author tag to `_npmUser` as that is the reliable one that npm populates. The `author` tag is developer populated, and can be missing.


src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -215,11 +216,14 @@ public override async Task<PackageMetadata> GetPackageMetadata(PackageURL purl)
                     metadata.VersionUri = $"{ENV_NPM_ENDPOINT}/package/{metadata.Name}/v/{metadata.PackageVersion}";
                     metadata.ApiVersionUri = $"{ENV_NPM_API_ENDPOINT}/{metadata.Name}/{metadata.PackageVersion}";
 
+                    // sometimes the description is only on the version level
+                    metadata.Description ??= OssUtilities.GetJSONPropertyStringIfExists(versionElement, "description");
```
brbayes-msft, line 220: nit: might it make more sense to prioritize the version specific description over the package one if it exists on the version?

pmalmsten:
What do we have in the way of unit tests / integration tests for this code? Should we add some kind of automated test demonstrating that this change works as intended now?

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -280,7 +287,7 @@ public override async Task<PackageMetadata> GetPackageMetadata(PackageURL purl)
                     }
 
                     // author(s)
-                    JsonElement? authorElement = OssUtilities.GetJSONPropertyIfExists(versionElement, "author");
+                    JsonElement? authorElement = OssUtilities.GetJSONPropertyIfExists(versionElement, "_npmUser");
```
pmalmsten, line 290: I see that both `author` (at the top level for a package) and `_npmUser` (on each version) exist in the package metadata: https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md

Could you elaborate a little on why "_npmUser" on the latest version is the preferred value to use? This is definitely better than the old code (the spec doesn't show an author element as existing on version metadata entries), but I'm curious about the semantics of the author for the package vs. the author for the version, if we should be recording both vs. just the author of the latest version, etc. etc.

jpinz: `_npmUser` is populated by the npm cli, it is the user that actually pushed the given version to the npm registry. `author` is populated in the `package.json` and could just be left empty. We could have both if we want to, and `_npmUser` is stored as a new `metadata.Publisher` value.

gfs:
> What do we have in the way of unit tests / integration tests for this code? Should we add some kind of automated test demonstrating that this change works as intended now?

I don't believe there is currently
Much in the metadata tests. I think this would be a good opportunity to add a few at least for npm.

jpinz:
> What do we have in the way of unit tests / integration tests for this code? Should we add some kind of automated test demonstrating that this change works as intended now?

Just added some tests, I added the json files as well, but I don't know how this testing framework does fakes so it currently uses the live HTTP client to get the metadata.
Hopefully we can get this in soon and then come back to replacing the live client with the json files, trying to get this in to unblock our provenance check.

gfs:
The existing tests should be mstest. 

One resource I found for mocking http client factory:
https://www.thecodebuzz.com/mock-typed-httpclient-httpclientfactory-moq-net-core/

gfs:
LGTM. Let me know if you're ready to merge with this. @jpinz

jpinz:
@gfs I think this should be good to go.

PR #307, opened 2022-03-09
Added a static OssGadgetJsonSerializer class

jpinz:
The new OssGadgetJsonSerializer contains the serialzer options needed for serializing and deserializing PackageURL objects.

Also fixed some issues with scoped npm packages and find squat mutations.

### DIFF
--- a/src/Shared/Extensions/PackageUrlExtension.cs
+++ b/src/Shared/Extensions/PackageUrlExtension.cs
@@ -2,9 +2,12 @@
 
 namespace Microsoft.CST.OpenSource.Extensions;
 
+using Helpers;
 using System.IO;
 using System.Text.RegularExpressions;
 using PackageUrl;
+using System.Net;
+using System.Web;
 
 public static class PackageUrlExtension
 {
@@ -12,5 +15,12 @@ public static class PackageUrlExtension
     {
         string invalidChars = new string(Path.GetInvalidFileNameChars()) + new string(Path.GetInvalidPathChars());
         return Regex.Replace(packageUrl.ToString(), "[" + Regex.Escape(invalidChars) + "]", "-");
+    }
+    
+    public static string GetNameWithScope(this PackageURL packageUrl)
+    {
+        return packageUrl.Namespace.IsNotBlank()
+            ? $"@{packageUrl.Namespace}/{packageUrl.Name}"
+            : packageUrl.Name;
     }
 }
--- a/src/Shared/Extensions/PackageUrlExtension.cs
+++ b/src/Shared/Extensions/PackageUrlExtension.cs
@@ -6,8 +6,6 @@ using Helpers;
 using System.IO;
 using System.Text.RegularExpressions;
 using PackageUrl;
-using System.Net;
-using System.Web;
 
 public static class PackageUrlExtension
 {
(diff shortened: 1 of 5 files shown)


src/Shared/Extensions/PackageUrlExtension.cs
```diff
@@ -2,9 +2,12 @@
 
 namespace Microsoft.CST.OpenSource.Extensions;
 
+using Helpers;
 using System.IO;
 using System.Text.RegularExpressions;
 using PackageUrl;
+using System.Net;
```
brbayes-msft, line 9: nit: Are both of these usings necessary?  I'm not sure as to what code might need them.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -105,22 +105,25 @@ public static class MutateExtension
 
             // Check to see if it is a scoped npm package to generate candidates for.
             bool isScoped = purl.Namespace.IsNotBlank() && purl.Type.Equals("npm", StringComparison.OrdinalIgnoreCase);
-            string nameToMutate = isScoped ? (purl.Namespace!).Substring(1) : purl.Name;
+            string nameToMutate = isScoped ? purl.Namespace : purl.Name;
```
brbayes-msft, line 108: Why is this change necessary?

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -105,22 +105,25 @@ public static class MutateExtension
...
                 foreach (Mutation mutation in mutator.Generate(nameToMutate))
                 {
                     // Construct the mutated name if the package was scoped.
-                    string mutated = isScoped ? $"@{mutation.Mutated}/{purl.Name}" : mutation.Mutated;
+                    string mutated = isScoped ? $"{mutation.Mutated}/{purl.Name}" : mutation.Mutated;
+                    string original = isScoped ? $"{purl.Namespace}/{purl.Name}" : purl.Name;
 
+                    Mutation calculatedMutation = new(mutated, original, mutation.Reason, mutation.Mutator);
+                    
                     if (generatedMutations.ContainsKey(mutated))
```
brbayes-msft, line 120: nit: this code can be simplified by having an if statement that checks if the dictionary entry exists, adding it if it does not, and then adding to that list outside of the if statement.

src/oss-tests/FindSquatsTests.cs
```diff
@@ -33,9 +34,9 @@ public async Task DetectSquats(string packageUrl, bool expectedToHaveSquats)
         }
 
         [DataTestMethod]
-        [DataRow("pkg:npm/%40angular/core", "@engular/core", "@angullar/core")]
+        [DataRow("pkg:npm/angular/core", "engular/core", "angullar/core")]
         [DataRow("pkg:npm/lodash", "odash", "lodah")]
-        [DataRow("pkg:npm/%40babel/runtime", "@abel/runtime", "@bable/runtime")]
+        [DataRow("pkg:npm/babel/runtime", "abel/runtime", "bable/runtime")]
```
brbayes-msft, line 39: Any potential of breaking backwards compatibility for users moving from older versions of the library to this update?

brbayes-msft (approved):

src/oss-find-squats-lib/MutateExtension.cs
gfs: This change will no longer remove the @ from the front of the namespace.  Won't that break the mutations?

src/oss-tests/FindSquatsTests.cs
gfs: Did the change to the packageurl library upstream implicitly change how we handle @'s in scopes?

gfs (approved):

src/oss-find-squats-lib/MutateExtension.cs
jpinz: PackageUrl stores the namespaces without the @ so for `@angular/core`, `nameToMutate` was being set as `ngular`

src/oss-tests/FindSquatsTests.cs
jpinz: Yeah, we drop the @ in packageurl now, it's just stored in the FindPackageSquatsResult as the name now I think.

jpinz: Back compat still works, just added:

`[DataRow("pkg:npm/%40angular/core", "@engular/core", "@angullar/core")]`

works fine

gfs: LGTM then.

brbayes-msft (approved):

PR #308, opened 2022-03-10
Fixed some bugs I found when trying to integrate

jpinz:
Some mutators were broken:
- No longer generates mutations with a new `/` at the beginning, so no `pkg:npm//lodash` (DoubleHit), no `pkg:npm//odash` (CloseLetters)
- No longer generates mutations with the same letters being swapped, mutating `pkg:npm/foo` and swapping the `'o'` generates the same package name, but was considered a mutation.

Fixed some bugs I found with typo squat generation, was broken for namespaces,


src/Shared/Extensions/PackageUrlExtension.cs
```diff
@@ -32,6 +33,15 @@ public static bool HasNamespace(this PackageURL packageUrl)
     /// <returns>The full name.</returns>
     public static string GetFullName(this PackageURL packageUrl)
     {
-        return packageUrl.HasNamespace() ? $"{packageUrl.Namespace}/{packageUrl.Name}" : packageUrl.Name;
+        if (!packageUrl.HasNamespace())
+        {
+            return packageUrl.Name;
+        }
+
+        string? namespaceStr = packageUrl.Type.Equals("npm", StringComparison.OrdinalIgnoreCase)
```
brbayes-msft, line 41: Please add code comments to explain why this is necessary.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -24,8 +24,7 @@ public class NuGetProjectManager : BaseProjectManager
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0044:Add readonly modifier", Justification = "Modified through reflection.")]
         public static string ENV_NUGET_ENDPOINT_API = "https://api.nuget.org";
         public static string ENV_NUGET_ENDPOINT = "https://www.nuget.org";
-
-        private readonly string NUGET_DEFAULT_REGISTRATION_ENDPOINT = "https://api.nuget.org/v3/registration5-gz-semver2/";
+        public static string NUGET_DEFAULT_REGISTRATION_ENDPOINT = "https://api.nuget.org/v3/registration5-gz-semver2/";
```
brbayes-msft, line 27: This seems like it should remain readonly.  Alternatively, you might be able to make it const.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -113,15 +113,17 @@ public static class MutateExtension
                 foreach (Mutation mutation in mutator.Generate(nameToMutate))
                 {
                     // Construct the mutated name if the package was scoped.
-                    string mutated = isScoped ? $"{mutation.Mutated}/{purl.Name}" : mutation.Mutated;
-                    string original = isScoped ? $"{purl.Namespace}/{purl.Name}" : purl.Name;
+                    string? newNamespace = isScoped ? mutation.Mutated : null;
```
brbayes-msft, line 116: How about mutatedNamespace and mutatedName instead?

brbayes-msft (changes requested):

brbayes-msft (approved):

pmalmsten (approved):

brbayes-msft (approved):

pmalmsten (approved):

PR #313, opened 2022-03-16
Migrate NuGetProjectManager to use the NuGet.Protocol library where applicable.

jpinz:
I was running into miscellaneous issues with parsing the JSON responses from the NuGet API, and after doing some digging around I found that the [Nuget.Client SDK provides a "Get package metadata" method](https://docs.microsoft.com/en-us/nuget/reference/nuget-client-sdk#get-package-metadata).
I went through and updated most of the methods in the NuGetProjectManager to use the library where I could.

- Downloading a package now uses the library.
- Enumerating through the list of versions now uses the library.
- Checking for existence of a package uses the library.
- Package Metadata now uses the library.


jpinz:
I'm logging off for dinner, tomorrow I will address any comments. I also plan on adding some more documentation and tests.

iarna:
Broadly the changes make sense, but I have a few git hygiene nit picks, all IMO of course:

The PR should have a name representing all of the PR (and not just a single commit)

The "[Clean up the NuGetProjectManager after I made a bunch of changes.](https://github.com/microsoft/OSSGadget/pull/313/commits/f10bcf94f0e4a06c5b2ddad458a8fe4e4a557b3c)" commit has semantic changes and those should be split out.

I'd split the npm changes out into their own commit. Generally a PR should do one thing.

gfs (changes requested):
Seems like a good idea. The new implementation of nuget manager is definitely simpler. However, the nuget download tests are failing so cannot approve as is.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -84,16 +92,16 @@ private async Task<string> GetRegistrationEndpointAsync()
         }
 
         /// <summary>
-        ///     Download one NuGet package and extract it to the target directory.
+        /// Download one NuGet package and extract it to the target directory.
         /// </summary>
         /// <param name="purl"> Package URL of the package to download. </param>
```
bhuwan-agarwal, line 97: Looks like the comments here don't reflect all the arguments. Maybe also add clarification on how the target directory is selected since it isn't being provided by the caller.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -104,93 +112,45 @@ public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl,
...
-                                {
-                                    targetName += Path.GetExtension(archive) ?? "";
-                                    await File.WriteAllBytesAsync(targetName, await result.Content.ReadAsByteArrayAsync());
-                                    downloadedPaths.Add(targetName);
-                                }
-                                return downloadedPaths;
-                            }
-                        }
+                
+                if (doExtract)
```
bhuwan-agarwal, line 136: Trying to understand the purpose of `doExtract` flag here as in when and how it's used by the caller.

If it's a much needed argument can't be removed, this if/else block can potentially be combined with the above if block where `doExtract` is true as well.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -308,26 +226,6 @@ public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL pur
                 return null;
```
bhuwan-agarwal, line 226: Should this method be updated to use API from the Nuget library as well?

src/Shared/PackageManagers/NuGetProjectManager.cs
pmalmsten: Looks like the prior `if` statement block has additional conditions beyond just `doExtract`, so I'm not sure these can be directly merged @bhuwan-agarwal

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +235,241 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
 
-        public JsonElement GetLatestCatalogEntry(JsonElement root)
-        {
-            return root.GetProperty("items").EnumerateArray().Last() // Last CatalogPage
-                .GetProperty("items").EnumerateArray().Last() // Last Package
-                .GetProperty("catalogEntry"); // Get the entry for the most recent package version
-        }
-        
-        public override JsonElement? GetVersionElement(JsonDocument? contentJSON, Version desiredVersion)
+        private async Task<JsonElement?> GetCatalogEntry(JsonDocument doc, PackageURL purl)
```
bhuwan-agarwal, line 357: Do this method and the ones below have any purpose anymore?

src/Shared/PackageManagers/NuGetProjectManager.cs
gfs: It looks to me like the second `if (doExtract)` maybe should be an else if to first.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -104,93 +112,45 @@ public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl,
...
+                CancellationToken cancellationToken = CancellationToken.None;
+
+                FindPackageByIdResource resource = await _sourceRepository.GetResourceAsync<FindPackageByIdResource>();
+
+                PackageIdentity packageIdentity = new(purl.Name, NuGetVersion.Parse(purl.Version));
+
+                IPackageDownloader? downloader = await resource.GetPackageDownloaderAsync(
+                    packageIdentity,
+                    _sourceCacheContext,
+                    NullLogger.Instance,
```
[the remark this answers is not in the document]
gfs: I don't believe that there is a mechanism to pass a logger, but I think it would be fine to add an ILoggerFactory parameter and convert the library to use ILogger. (in a separate PR for changing this across the project).

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -262,6 +262,24 @@ public static IEnumerable<string> SortVersions(IEnumerable<string> versionList)
...
+        {
+            List<KeyValuePair<string, DateTime>> versionList = versionDictionary.ToList();
+
+            versionList.Sort((pair1, pair2) => pair2.Value.CompareTo(pair1.Value));
+            if (!versionList.Any())
+            {
+                return Array.Empty<string>();
+            }
+            
+            return versionList.Select(kvp => kvp.Key).Distinct();
```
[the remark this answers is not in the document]
jpinz: We want it in descending order, so just using OrderByDescending. Addressed in my other branch that I cherry-picked this code out to. It wasn't related to any of the NuGet changes.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -112,20 +112,48 @@ public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL pur
...
                 }
-                return SortVersions(versionList.Distinct());
+
+                foreach ((string version, _) in versionDictionary)
+                {
+                    if (!versionList.Contains(version))
+                    {
+                        versionDictionary.Remove(version);
+                    }
+                }
```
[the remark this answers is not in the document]
gfs: I think the simplest thing to do instead of this loop would be:

`versionDictionary = versionDictionary.Where((version, _) => versionList.Contains(key));`

src/Shared/PackageManagers/NuGetProjectManager.cs
pmalmsten: One option might be to move the `Directory.Exists(extractionPath) && cached` portion of the condition into a nested if statement here

jpinz:
@iarna @bhuwan-agarwal Cherry-Picked the NPM changes into their own PR: #314

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -104,93 +112,45 @@ public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl,
...
-                                return downloadedPaths;
-                            }
-                        }
+                
+                if (doExtract)
+                {
+                    IEnumerable<string>? files = await downloader.CoreReader.GetFilesAsync(extractionPath, cancellationToken);
+                    if (files is not null && files.Any())
+                    {
+                        downloadedPaths.Add(extractionPath);
```
pmalmsten: I like that after these changes, this method is much smaller, but I wonder whether we can do even better - what do you think about splitting the extract and non-extract into two separate private helper methods, which this method can call? Might help make the high-level logic easier to read, rather than having long method bodies.

src/Shared/PackageManagers/NuGetProjectManager.cs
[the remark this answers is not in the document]
jpinz: Yeah, I might file an issue asking if they can just migrate over to the official implementation now. Any thoughts on that?

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: From poking around and testing, it's basically do you want to just download the `.nupkg` or do you want to download the `.nupkg` but extract the contents into a directory, leaving you with something like this:
image

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: I can do that!

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: Updating it to use the NuGet.Protocol library.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: No, removed.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: Actually, @pmalmsten don't know what version of the PR you were reviewing when looking at this, but the logic for `doExtract` and the else is now only 10 lines long. Let me know if you think it's still worth refactoring it.

src/Shared/PackageManagers/NuGetProjectManager.cs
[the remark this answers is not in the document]
jpinz: The NuGet.Protocol library actually defines their own ILogger, it isn't from Microsoft.Extensions.Logging from what I can tell. https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Common/Logging/ILogger.cs

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +235,241 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
         /// <inheritdoc />
         public override async Task<PackageMetadata> GetPackageMetadata(PackageURL purl, bool useCache = true)
```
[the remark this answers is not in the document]
jpinz: Done! I tried to separate it where it made sense.

gfs:
Moq looks like it could work for the nuget protocol tests:

https://github.com/moq/moq4

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: I wanted these two statements separated because if the first if statement, `if (doExtract && Directory.Exists(extractionPath) && cached)` evaluates to true, then the `resource.CopyNupkgToStreamAsync` doesn't need to be ran and we can return results earlier. It might not be a huge deal, but that's just how I was looking at it.

src/Shared/PackageManagers/NuGetProjectManager.cs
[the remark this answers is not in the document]
gfs: I just attempted to make this wrapper but it is not possible because the ILogger they does not have public members.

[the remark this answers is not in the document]
jpinz: Added a wrapper @gfs.

jpinz:
Fixed some unit test failures. Logging off but will work with moq tomorrow to fix the unit tests using live data.

jpinz:
I reworked the NuGetProjectManagerTests to use `moq` and mocked implementations. So those tests no longer use live data. Going to rework the rest of the relevant tests on Monday.

jpinz:
I think this PR should be good to review again. I will be publishing a subsequent PR, probably as a draft that does a lot of refactoring for adding an `IManagerProvider` and `IManagerMetadata` instead of the provider for strictly NuGet.
Before the PR is out, you can view the branch here: https://github.com/jpinz/OSSGadget/tree/jupinzer/projectmanager_provider

src/Shared/PackageManagers/PackageManagerFactory.cs
```diff
@@ -76,6 +77,57 @@ public static BaseProjectManager CreateBaseProjectManager(string destinationDire
 
             return null;
         }
+        
+        /// <summary>
+        /// Get an appropriate project manager for package given its PackageURL, with a provider.
+        /// </summary>
+        /// <param name="purl">The <see cref="PackageURL"/> for the package to create the project manager for.</param>
+        /// <param name="nuGetProvider">The <see cref="INuGetProvider"/> for to make calls with.</param>
```
gfs, line 85: Is this a general purpose interface?  The comments indicate it is but the types indicate it is not.

If this is just for Nuget the API name + comments should indicate that and it not the INugetProvider should be refactored to inherit from a IProvider interface + perhaps rename the "GetNuPkg" method to get the package or move that call to the nuget specific interface.

src/Shared/PackageManagers/PackageManagerFactory.cs
```diff
@@ -76,6 +77,57 @@ public static BaseProjectManager CreateBaseProjectManager(string destinationDire
...
+            }
+            // Use reflection to find the correct package management class
+            Type? downloaderClass = projectManagers
+               .Where(type => type.Name.Equals($"{purl.Type}ProjectManager",
+                                               StringComparison.InvariantCultureIgnoreCase))
+               .FirstOrDefault();
+
+            if (downloaderClass != null)
+            {
+                if (downloaderClass != typeof(NuGetProjectManager))
```
gfs, line 104: If its not nuget we just shortcut this method and ignore the provider? This doesn't seem correct to me.

src/oss-metadata/MetadataTool.cs
```diff
@@ -55,7 +56,7 @@ public MetadataTool(IHttpClientFactory httpClientFactory) : base(httpClientFacto
             try
             {
                 // Use reflection to find the correct downloader class
-                BaseProjectManager? projectManager = ProjectManagerFactory.CreateProjectManager(purl, httpClientFactory);
+                BaseProjectManager? projectManager = ProjectManagerFactory.CreateProjectManagerWithProvider(purl, new NuGetProvider(), httpClientFactory);
```
gfs, line 59: This change is very messy.  If CreateManagerWithProvider always takes a `new NuGetProvider()` with no parameters why not just construct it in the old create project manager method only when needed?

gfs (changes requested):

src/Shared/PackageManagers/PackageManagerFactory.cs
gfs: I see why you are doing this now based on the way you are calling it but I don't think this is the correct way.

src/oss-metadata/MetadataTool.cs
jpinz: This was a quick first try solution for me. I have already refactored a lot of this into a new branch. But the change is pretty large scoped so I wasn't sure if I should include it with this PR https://github.com/jpinz/OSSGadget/tree/jupinzer/projectmanager_provider

gfs: This other version looks much better.  I think it makes more sense to do it holistically in one PR unless you need the nuget specific version merged imminently for some reason.  It looks like the new interface declared in this PR are changed right away with the other branch.

jpinz: I published the other PR as a draft if you want to check it out.

src/Shared/Model/NuGetMetadata.cs
```diff
@@ -0,0 +1,167 @@
...
+using NuGet.Versioning;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Threading.Tasks;
+
+/// <summary>
+/// A class to represent Package Metadata for a NuGet package.
+/// </summary>
+public class NuGetMetadata : IPackageSearchMetadata
```
pmalmsten, line 19: Nit: [immutable record types](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record) offer some nice properties - do you think this could be made an immutable record class?

src/Shared/Model/NuGetMetadata.cs
```diff
@@ -0,0 +1,167 @@
...
+    
+    [JsonProperty(PropertyName = JsonProperties.SubjectId)]
+    public Uri CatalogUri { get; private set; }
+
+    // Constructor for json deserialization
+    [JsonConstructor]
+#pragma warning disable CS8618
+    public NuGetMetadata()
+#pragma warning restore CS8618
+    {}
```
pmalmsten, line 115: This seems a bit odd - why bother defining a constructor with `[JsonConstructor]` when it accepts no arguments? Could we simply delete these lines?

src/Shared/Model/PackageMetadata.cs
```diff
@@ -177,6 +178,17 @@ public class User
...
+        protected bool Equals(User other) => Active == other.Active && Email == other.Email && Id == other.Id && Name == other.Name && Url == other.Url;
+
+        public override bool Equals(object? obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            return obj.GetType() == this.GetType() && Equals((User) obj);
+        }
+
+        public override int GetHashCode() => HashCode.Combine(Active, Email, Id, Name, Url);
```
pmalmsten, line 191: Nit: Any chance that this class could be made an [immutable record type](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record) instead? We might be able to get these definitions for free.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -208,77 +144,26 @@ public override async Task<bool> PackageExists(PackageURL purl, bool useCache =
...
-                        }
-                        else
-                        {
-                            Logger.Debug("Catalog identifier was null.");
-                        }
-                    }
-                }
-                return SortVersions(versionList.Distinct());
+                // Sort versions, highest first, lowest last.
+                return SortVersions(versions);
```
pmalmsten, line 166: It looks like this sorting behavior is not documented on the parent type. Should this ordering be documented on the parent type? Do other implementations of this method for the parent type also follow this same ordering

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -292,39 +177,31 @@ public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL pur
...
-            try
-            {
-                HttpClient httpClient = CreateHttpClient();
 
-                string? content = await GetHttpStringCache(httpClient, versionUri, useCache);
-                if (string.IsNullOrEmpty(content)) { return null; }
+                // If no package version provided, default to the latest version.
+                if (string.IsNullOrWhiteSpace(packageVersion))
+                {
+                    string latestVersion = (await EnumerateVersions(purl, useCache)).First();
```
pmalmsten, line 190: I see this latest version pattern in lots of places in this file - what do you think about a private helper method called `LatestVersion`? Maybe that functionality is useful enough to raise it into the parent type?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +214,210 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
-            metadata.Description = latestCatalogEntry.GetProperty("description").GetString();
-            
-            // Title is different than name or description for NuGet packages, not always used.
-            // metadata.Title = GetLatestCatalogEntry(root).GetProperty("title").GetString();
+            NuGetMetadata? packageVersion =
+                await _nuGetProvider.GetMetadataAsync(purlWithVersion, useCache: useCache);
+
+            if (packageVersion is null)
+            {
+                throw new NullReferenceException();
```
pmalmsten, line 228: Nit: `NullReferenceException` is typically [reserved for use by the runtime](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types#nullreferenceexception-indexoutofrangeexception-and-accessviolationexception). How about `InvalidOperationException` instead?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +214,210 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
+            }
```
pmalmsten, line 229: Nit: you can compress null checks like this using the null coalescing operator `??`:

```suggestion
            NuGetMetadata packageVersion =
                await _nuGetProvider.GetMetadataAsync(purlWithVersion, useCache: useCache) ?? throw new InvalidOperationException("Could not get metadata for {PURL}", purlWithVersion);
```

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +214,210 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
 
-            metadata.PackageManagerUri = ENV_NUGET_ENDPOINT;
+            metadata.PackageManagerUri = ENV_NUGET_ENDPOINT_API;
             metadata.Platform = "NUGET";
             metadata.Language = "C#";
-            // metadata.SpokenLanguage = GetLatestCatalogEntry(root).GetProperty("language").GetString();
-            metadata.PackageUri = $"{metadata.PackageManagerUri}/packages/{metadata.Name?.ToLower()}";
-            metadata.ApiPackageUri = $"{RegistrationEndpoint}{metadata.Name?.ToLower()}/index.json";
+            metadata.PackageUri = $"{ENV_NUGET_HOMEPAGE}/{metadata.Name?.ToLowerInvariant()}";
+            metadata.ApiPackageUri = $"{RegistrationEndpoint}{metadata.Name?.ToLowerInvariant()}/index.json";
```
pmalmsten, line 240: Can we actually tolerate `null` for `metadata.Name` here? Or would these property values just be garbage if that was the case?

If these would be garbage, would it make sense to use a non-nullable value (maybe `packageVersion.PackageId`), or to assert that the value is not null?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +214,210 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
+                return;
+            }
 
-                    // TODO: maintainers
+            string? nameLowercase = metadata.Name?.ToLowerInvariant();
 
-                    // repository
-                    PackageURL nuspecPurl = purl;
+            // Set the version specific URI values.
+            metadata.VersionUri = $"{metadata.PackageManagerUri}/packages/{nameLowercase}/{metadata.PackageVersion}";
```
pmalmsten, line 266: Can we actually tolerate `nameLowercase` being null? Or would the resulting `VersionUri` value just be garbage?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +214,210 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
+            metadata.ApiVersionUri = packageVersion.CatalogUri.ToString();
+            
+            // Construct the artifact contents url.
+            metadata.VersionDownloadUri = GetNupkgUrl(metadata.Name!, metadata.PackageVersion);
```
pmalmsten, line 270: The fact that we are using `metadata.Name!` here also suggests that we should be actually checking whether `metadata.Name` is null or not earlier (see my above comment)

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +214,210 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
-                        Name = OssUtilities.GetJSONPropertyStringIfExists(versionElement, "licenseExpression"),
-                        Url = OssUtilities.GetJSONPropertyStringIfExists(versionElement, "licenseUrl")
-                    });
-                    
-                    // publishing info
-                    metadata.UploadTime = OssUtilities.GetJSONPropertyStringIfExists(versionElement, "published");
+                foreach (PackageDependencyGroup dependencyGroup in dependencyGroups)
+                {
+                    dependencyGroup.Packages.ToList().ForEach((dependency) => metadata.Dependencies.Add(new Dependency() { Package = dependency.ToString(), Framework = dependencyGroup.TargetFramework?.ToString()}));
                 }
```
pmalmsten, line 292: Nit: I wonder whether this could be simplified a bit:

```c#
metadata.Dependencies ??= dependencyGroups.SelectMany(group => group.Packages, (dependencyGroup, package) => new { dependencyGroup, package})
                                   .Select(dependencyGroupAndPackage => new Dependency() { Package = dependencyGroupAndPackage.package.ToString(), Framework = dependencyGroupAndPackage.dependencyGroup.TargetFramework?.ToString()})
                                   .ToList();
```

Or are we expecting `metadata.Dependencies` to possibly already have entries we want to keep?

pmalmsten (changes requested):
Got through as much as I could for today, will keep looking tomorrow.

src/Shared.CLI/Helpers/PackageDownloader.cs
```diff
@@ -47,7 +48,7 @@ public PackageDownloader(PackageURL purl, IHttpClientFactory? httpClientFactory,
                 destinationDirectory = destinationDir;
             }
 
-            packageManager = ProjectManagerFactory.CreateProjectManager(purl, httpClientFactory, destinationDirectory);
+            packageManager = ProjectManagerFactory.CreateProjectManagerWithProvider(purl, new NuGetProvider(), httpClientFactory, destinationDirectory);
```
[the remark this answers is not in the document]
jpinz: This is being removed in the other PR I have out. I think this has been addressed there.

src/Shared/Contracts/INuGetProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// Downloads a .nupkg file for a given <see cref="PackageURL"/>, and optionally extracts it.
+    /// </summary>
+    /// <param name="projectManager">The <see cref="NuGetProjectManager"/> this method is being called from, to use for file extraction.</param>
+    /// <param name="packageUrl">The <see cref="PackageURL"/> to download the .nupkg of.</param>
+    /// <param name="targetDirectory">The directory to save the contents to.</param>
+    /// <param name="doExtract">If the contents of the .nupkg should be extracted.</param>
+    /// <param name="cached">If the downloaded contents should be retrieved from the cache if they exist there.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to be used in the method call, defaults to <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The location of the download, or null if the download failed.</returns>
+    Task<string?> DownloadNupkgAsync(NuGetProjectManager projectManager, PackageURL packageUrl, string targetDirectory, bool doExtract, bool cached = false, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
jpinz: In the new PR, I take in a `BaseProjectManager` instead. This is used because the `ExtractArchive` method is defined in the `BaseProjectManager` I would be ok with refactoring this somewhere else. Maybe not right now, but in the future. The only dependency `ExtractArchive` has in the class is the `TopLevelExtractionDirectory` which could easily be passed in as a parameter.

src/Shared/Model/NuGetMetadata.cs
jpinz: That is necessary for the way Newtonsoft.json constructs the class from json, I believe. I think I couldn't get it working otherwise. I will mess around with it tomorrow.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: They all follow this same ordering, with the exception (in another open PR) for NPM. Where the ordering still applies, but if the version tagged `latest` isn't the highest version according to the sort, it gets removed from it's position in the list and placed at the beginning of the list.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: I thought about doing that yes. I can mess around with it tomorrow.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: Resolving in other PR.

src/Shared/Contracts/INuGetProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+    Task<IEnumerable<string>> GetAllVersionsAsync(PackageURL packageUrl, bool useCache = true, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Gets the <see cref="PackageSearchMetadataRegistration"/> for this package version.
+    /// </summary>
+    /// <param name="packageUrl">The <see cref="PackageURL"/>. Requires a version.</param>
+    /// <param name="useCache">If the cache should be checked for the metadata of this package.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to be used in the method call, defaults to <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The <see cref="NuGetMetadata"/> for this package version.</returns>
+    /// <exception cref="ArgumentNullException">Thrown if there was no version in packageUrl</exception>
```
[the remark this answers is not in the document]
jpinz: Resolving in other PR.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: Resolved in other PR.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: Resolved in the other PR.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: No, and it should be fixed now. Resolved in the other PR.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: Resolved in other PR.

jpinz:
Refactored most of this work into a new PR #313. Will be addressing the rest of the comments, and moving anything not already refactored from this PR tomorrow.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: Resolved in new PR.

src/Shared/Model/NuGetMetadata.cs
jpinz: Resolved in new PR.

src/Shared/Model/PackageMetadata.cs
jpinz: Resolved in other PR.

src/Shared/Model/NuGetMetadata.cs
jpinz: I tried removing it and the unit tests from NuGetProjectManagerTests started failing. It has to do with the json parsing. I think I had seen this used elsewhere before but can't find it.

src/Shared/PackageManagers/PackageManagerFactory.cs
jpinz: I believe this should be resolved in the new PR.

src/Shared/PackageManagers/PackageManagerFactory.cs
jpinz: Also should be resolved in the new PR. I no longer do any checks for if it is a NuGetProjectManager or not.

src/oss-metadata/MetadataTool.cs
jpinz: The new PR is out.

PR #314, opened 2022-03-17
NPM Sort versions by SemVer, but always put the version tagged `latest` first.

jpinz:
I updated how enumerating versions of an NPM package works. It sorts them with SemVer like it was doing, but moves the version number tagged `latest` in `dist-tags.latest` from the registry as the first version in the list.


gfs:
This raises an interesting question to me.  In this case both of these are actually prerelease versions - so most users are not going to be using either of these versions even though they are the latest in their respective major releases.

jpinz:
> This raises an interesting question to me. In this case both of these are actually prerelease versions - so most users are not going to be using either of these versions even though they are the latest in their respective major releases.

Yeah, no clue what the right solution is here, it could be that we don't consider `latestVersion` to be a prerelease version, or a beta/alpha version. It could mean "whatever you get when you do `npm install packageName`"

Any thoughts @iarna?

scovetta:
When I think about "latest", I think about "what would `npm install foo` do?"

But for npm, this really means "whatever is tagged `latest`", which is defined under `dist-tags` in the registry. Other package managers could of course do it differently. I suppose we could use that for registries that provide it, and if not, sort by semver, and if ambiguous, sort by time?

gfs:
> When I think about "latest", I think about "what would `npm install foo` do?"
> 
> 
> 
> But for npm, this really means "whatever is tagged `latest`", which is defined under `dist-tags` in the registry. Other package managers could of course do it differently. I suppose we could use that for registries that provide it, and if not, sort by semver, and if ambiguous, sort by time?

I agree, and think the behavior should try to match the default
behavior of the original package manager tool. 

In general, even if the manager api we use doesn't set a prerelease flag, if it's set by semver (having a - flag in the version) that shouldn't come up as latest.

Maybe this API needs a bool flag to accept/not accept prereleases?

jpinz:
> Maybe this API needs a bool flag to accept/not accept prereleases?

I agree with this, do we want to try to accomplish that in this PR? Or just merge this one and make an issue for that feature enhancement?

gfs:
The Change to do so would need to be to the underlying interface right so it would require updating potentially all the managers.

It may make sense to do this in its own PR.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -112,20 +112,36 @@ public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL pur
...
+                versionList.Remove(latestVersion);
+                
+                // Sort the list of distinct versions.
+                List<string> sortedList = SortVersions(versionList.Distinct()).ToList();
+                
+                // Insert the latest version at the beginning of the list.
+                sortedList.Insert(0, latestVersion);
+
+                return sortedList;
+
```
gfs, line 144:
```suggestion
```
Nit extra line.

src/oss-tests/Properties/Resources.Designer.cs
```diff
@@ -61,7 +61,7 @@ internal class Resources {
...
+        ///   Looks up a localized string similar to {&quot;_id&quot;:&quot;@angular/core&quot;,&quot;_rev&quot;:&quot;660-3fe4d5b1b9e632c7d96cdc1e1005b13c&quot;,&quot;name&quot;:&quot;@angular/core&quot;,&quot;dist-tags&quot;:{&quot;latest&quot;:&quot;13.2.6&quot;,&quot;v4-lts&quot;:&quot;4.4.7&quot;,&quot;v5-lts&quot;:&quot;5.2.11&quot;,&quot;next&quot;:&quot;13.3.0-rc.0&quot;,&quot;v7-lts&quot;:&quot;7.2.15&quot;,&quot;v6-lts&quot;:&quot;6.1.10&quot;,&quot;v9-lts&quot;:&quot;9.1.13&quot;,&quot;v8-lts&quot;:&quot;8.2.14&quot;,&quot;v10-lts&quot;:&quot;10.2.5&quot;,&quot;v11-lts&quot;:&quot;11.2.14&quot;,&quot;v12-lts&quot;:&quot;12.2.16&quot;},&quot;versions&quot;:{&quot;0.0.0-0&quot;:{&quot;name&quot;:&quot;@angular/core&quot;,&quot;version&quot;:&quot;0.0.0-0&quot;,&quot;description&quot;:&quot;&quot;,&quot;main&quot;:&quot;index.js&quot;,&quot;jsnext:main&quot;:&quot;esm/index.js&quot;,&quot;typings&quot;:&quot;index.d.ts&quot;,&quot;author&quot;:{&quot;name&quot;:&quot;angular&quot;},&quot;license&quot;:&quot;MIT&quot;,&quot;pee [rest of string was tru....
```
gfs, line 64: Where is the resources.resx file that generated this?

jpinz: I was trying to find it, but I think it's because the resources.resx didn't change. The json files already existed, and the changes are due to that I changed the actual json file contents, so the Reources.Designer changed to reflect the contents of the json files changing.

gfs: Ah so the compilation extracts the contents of the referenced json and embeds it here?

gfs (approved):
I think this is fine now, but do make sure to follow-up with the larger refactor that respects a prerelease flag.

src/oss-tests/Properties/Resources.Designer.cs
jpinz: Pretty sure that's what it does, yes.

gfs: Since this file is generated should it even be checked in? or should it be gitignored?

jpinz: I don't think it gets auto generated. From the research I did, it needs to be checked in as well. https://stackoverflow.com/questions/69876372/why-cant-i-delete-designer-cs-files

PR #318, opened 2022-03-23
Generic IManagerProvider and IManagerMetadata

jpinz:
Implement an `IManagerProvider` and an `IManagerMetadata` interface that gets created by an `IManagerProviderFactory` implemented by `ProviderFactory`.
Eventually a lot of the logic for handling manager specific functionality can be split into it's own projects/libraries/repos/etc like the NuGet.Protocol does for NuGet.


src/Shared.CLI/Helpers/PackageDownloader.cs
```diff
@@ -23,9 +24,10 @@ public class PackageDownloader
         /// </summary>
         /// <param name="purl">The package to download.</param>
         /// <param name="httpClientFactory">The <see cref="IHttpClientFactory"/> to use to make the client to use when downloading.</param>
+        /// <param name="provider">The <see cref="IManagerProvider{IManagerMetadata}"/> to use.</param>
         /// <param name="destinationDir">The directory where the package needs to be downloaded to.</param>
         /// <param name="doCaching">Check and use the cache if it exists - create if not.</param>
-        public PackageDownloader(PackageURL purl, IHttpClientFactory? httpClientFactory, string? destinationDir = null, bool doCaching = false)
+        public PackageDownloader(PackageURL purl, IHttpClientFactory? httpClientFactory, IManagerProvider<IManagerMetadata>? provider = null, string? destinationDir = null, bool doCaching = false)
```
gfs, line 30: Do we really need both an IHttpClientFactory and a IManagerProvider? Should the requests not flow through the ManagerProvider (which might itself take an IHttpClientFactory depending on the implementation).

src/Shared/PackageManagers/PackageManagerFactory.cs
```diff
@@ -16,29 +18,30 @@ public static class ProjectManagerFactory
...
         /// <summary>
         /// Get an appropriate project manager for package given its PackageURL.
         /// </summary>
         /// <param name="purl">The <see cref="PackageURL"/> for the package to create the project manager for.</param>
         /// <param name="httpClientFactory"> The <see cref="IHttpClientFactory"/> for the project manager to use for making Http Clients to make web requests.</param>
+        /// <param name="managerProvider">The <see cref="IManagerProvider{IManagerMetadata}"/> for this manager.</param>
         /// <param name="destinationDirectory">The directory to use to store any downloaded packages.</param>
         /// <returns> BaseProjectManager object </returns>
-        public static BaseProjectManager? CreateProjectManager(PackageURL purl, IHttpClientFactory? httpClientFactory = null, string? destinationDirectory = null)
+        public static BaseProjectManager? CreateProjectManager(PackageURL purl, IHttpClientFactory? httpClientFactory = null, IManagerProvider<IManagerMetadata>? managerProvider = null, string? destinationDirectory = null)
```
gfs, line 44: I don't understand how this pattern where you provide the provider here is going to work. It means you have to know ahead of time which manager your purl associates with, which doesn't fit this API surface - which has always been designed to take an arbitrary purl.

gfs (changes requested):
I see a couple issues here.

1. I'm not sure why you would need to provide both an IHttpClient and an IManagerProvider. If the goal is to replace the amnager providers with individual implementations it should be the only parameter and the IHttpClientFactory should pass into the manager when needed.
2. I don't think these new proposed public APIs make sense. This is most obvious with the packagemanagerfactory but you can also see it with the tools. The CLI tools and createprojectmanager method are designed to take any arbitrary purl, but if you are providing them the ManagerProvider it will obviously only work with a matching purl.  I think this needs a rethink on how you get the manager provider.  Maybe it should be a manager provider factory that can produce the manager provider with your selected settings based on the purl?

src/Shared.CLI/Helpers/PackageDownloader.cs
jpinz: That's a good point. I hadn't considered putting the IHttpClientFactory in the provider. I will give that a shot.

src/Shared/PackageManagers/PackageManagerFactory.cs
jpinz: So my reasoning behind this, similarly to providing the IHttpClientFactory, is that most of the time a provider will not be provided. Just like most of the time we don't provide an IHttpClientFactory, except for our consumption case where we want to control the factory.
By providing the provider in the constructor, it allows us to provide a mocked implementation of the interface for unit test calls. There could very well be a better way of implementing this, but from my still somewhat limited understanding, we need to provide it here. 
I suppose one alternative is that I could add a method to the ProjectManagers that allows us to override and replace the provider that gets added during the factory construction (not from the parameters, but from the ProviderFactory) and then only use that method in unit tests?
If you know of any good design patterns for what I'm trying to accomplish please let me know. It really is only there for the unit test mocking, there shouldn't be a reason from any consuming level side for the consumer to want to override the provider like they might want to with the IHttpClientFactory.

gfs: What about this:

Add an `IManagerProviderFactory` which takes a purl and an `IHttpClientFactory` and produces the correct `IManagerProvider`.
Each `BaseProjectManager` should have an associated `IManagerProvider`.
Instead of passing an `IHttpClientFactory` pass the `IManagerProviderFactory`.
Optional: the `IManagerProvider` interface should expose the `IHttpClientFactory` that was provided to its factory when it was constructed as well.
When passed a null `IManagerProviderFactory` instantiate the default ones -  (keep same default behavior).
When running unit tests Mock the `IManagerProviderFactory` to produce `IManagerProvider` mocks.

gfs: This keeps the same flexibility as well to override the `IHttpClientFactory` as before without worrying about writing the `IManagerProvider` as the factory can pass in your `IHttpClientFactory` if you desire.

gfs: It looks like you may have started a bit with that on the `ProviderFactory.cs`. I think you should just fully commit to the `IProvider` pattern and replace the `HttpClientFactory` parameters entirely (on the managers), with that instead being (optionally) passed to the factory.

jpinz: Just made a huge refactor that did all of that. I think it accomplishes what you were thinking of.

gfs:
Will review refactor tomorrow.

src/Shared/Model/Providers/ProviderFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// Get an appropriate implementation of <see cref="BaseProvider"/> for the project manager in the provided
+    /// <see cref="PackageURL"/>. If no implementation of <see cref="BaseProvider"/> exists for this manager, returns null.
+    /// </summary>
+    /// <param name="purl">The <see cref="PackageURL"/> for the package to create the provider for.</param>
+    /// <returns>An implementation of <see cref="BaseProvider"/> for this project manager, or null.</returns>
+    public BaseProvider CreateProvider(PackageURL purl)
+    {
+        if (ManagerProviders.Count == 0)
+        {
+            ManagerProviders.AddRange(typeof(BaseProvider).Assembly.GetTypes()
```
brbayes-msft, line 42: Could use another reviewer here to double check that this isn't at risk for malicious implementations being injected at runtime.  I think it's okay because it's limited to the same assembly as BaseProvider, but could use someone else to double check.

src/Shared/PackageManagers/PackageManagerFactory.cs
```diff
@@ -28,50 +30,39 @@ public static BaseProjectManager CreateBaseProjectManager(IHttpClientFactory htt
         /// <returns></returns>
         public static BaseProjectManager CreateBaseProjectManager(string destinationDirectory)
         {
-            return new BaseProjectManager(destinationDirectory);
+            return new BaseProjectManager(new BaseProvider(), destinationDirectory);
```
brbayes-msft, line 33: If the intent here is to have a provider that complies with the interface but does not actually return any data, perhaps either a) the BaseProjectManager could be modified to allow the provider to be nullable, returning NotSupported if any relevant calls are made, or b) a NoOpProvider could be added instead of relying on functionality in a base class that does the same thing as the base class right now, but can instead be sealed so that no other classes try to extend it.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -84,16 +83,19 @@ private async Task<string> GetRegistrationEndpointAsync()
...
+        /// <param name="cached">If the downloaded contents should be retrieved from the cache if they exist there.</param>
+        /// <returns>An <see cref="IEnumerable{String}"/> list of the path(s) the contents were downloaded to.</returns>
         public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl, bool doExtract, bool cached = false)
         {
-            Logger.Trace("DownloadVersion {0}", purl?.ToString());
+            Logger.Trace("DownloadVersion {0}", purl.ToString());
 
-            string? packageName = purl?.Name;
-            string? packageVersion = purl?.Version;
+            string? packageName = purl.Name;
```
brbayes-msft, line 97: Is the PackageUrl no longer nullable here?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -208,77 +140,26 @@ public override async Task<bool> PackageExists(PackageURL purl, bool useCache =
...
         }
 
         /// <inheritdoc />
         public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL purl, bool useCache = true)
         {
-            Logger.Trace("EnumerateVersions {0}", purl?.ToString());
+            Logger.Trace("EnumerateVersions {0}", purl.ToString());
 
-            if (purl == null || purl.Name is null)
+            if (purl.Name is null)
```
brbayes-msft, line 152: Shouldn't this be `purl?.Name is null`?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +210,199 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
-                }
+                metadata.Authors ??= new List<User>();
+                authors.Split(", ").ToList()
+                    .ForEach(author => metadata.Authors.Add(new User() { Name = author }));
             }
-            catch (KeyNotFoundException) { return null; }
-            catch (InvalidOperationException) { return null; }
 
-            return null;
+            // TODO: Maintainers
```
brbayes-msft, line 314: Please write out this todo as a sentence.

brbayes-msft (changes requested):

src/Shared/Contracts/IManagerProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+/// <summary>
+/// The interface to implement project manager specific methods.
+/// Useful for mocking when doing unit tests.
+/// </summary>
+public interface IManagerProvider<T>
+{
+    /// <summary>
+    /// The <see cref="IHttpClientFactory"/> for this <see cref="IManagerProvider{T}"/> to make <see cref="HttpClient"/>s with.
+    /// </summary>
+    public IHttpClientFactory HttpClientFactory { get; }
```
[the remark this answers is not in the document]
gfs: Since many of the managers may use HttpClientFactory behind the scenes I think it makes sense to have an `IHttpClientBasedManagerProvider` or `BaseHttpClientBasedManagerProvider` that does have this field to make it easier to develop those. But agree the base interface does not need it.

src/Shared/Model/Providers/ProviderFactory.cs
gfs, line 42: I am not a fan of this implicit enumeration - I think it would make sense to have the list of available providers set before you call this method - it seems more appropriate for the constructor - perhaps also as an optional parameter to skip this reflection.

src/Shared/Contracts/IManagerProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+/// <summary>
+/// The interface to implement project manager specific methods.
+/// Useful for mocking when doing unit tests.
+/// </summary>
+public interface IManagerProvider<T>
```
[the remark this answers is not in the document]
gfs: Agree, I don't think this needs to be generic.

src/Shared/Model/Metadata/NuGetProtocolMetadata.cs
```diff
@@ -0,0 +1,166 @@
...
+        LicenseMetadata = registration.LicenseMetadata;
+        Vulnerabilities = registration.Vulnerabilities;
+        CatalogUri = registration.CatalogUri;
+    }
+
+    /// <summary>
+    /// Serialize this instance of <see cref="NuGetMetadata"/> into a json string.
+    /// </summary>
+    /// <returns>A json string representing this object.</returns>
+    public string ToJson()
```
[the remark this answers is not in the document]
gfs: Agree, I don't think this is needed. There are no special serialization options being used.

src/Shared/Model/Providers/NuGetProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+    /// Instantiates a new <see cref="NuGetProvider"/>.
+    /// </summary>
+    public NuGetProvider(IHttpClientFactory? httpClientFactory = null)
+    {
+        HttpClientFactory = httpClientFactory ?? new DefaultHttpClientFactory();
+    }
+
+    /// <inheritdoc />
+    public override async Task<string?> DownloadAsync(BaseProjectManager projectManager, PackageURL packageUrl, string targetDirectory, bool doExtract,
+        bool cached = false, CancellationToken cancellationToken = default)
```
gfs, line 39: I don't think this should take a project manager as an argument.  It doesn't appear to actually need the manager.

The only think it used for is `ExtractArchive` which is actually a method implemented on the base class as far as I can tell which just leverages `RecursiveExtractor` and could be refactored into a static helper method.

src/Shared/Model/Providers/ProviderFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            BaseProvider? provider = (BaseProvider)ctor.Invoke(new object?[] { HttpClientFactory });
+            return provider;
+        }
+
+        return new BaseProvider(HttpClientFactory);
+    }
+
+    // do reflection only once
+    private static readonly List<Type> ManagerProviders = new();
```
gfs, line 67: This reflection is actually done once per provider factory. It may make sense to do this reflection one time and pass the providers you want in instead when constructing the provider factory.

src/Shared/Model/Providers/ProviderFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+        if (providerClass == null)
+        {
+            return new BaseProvider(HttpClientFactory);
+        }
+
+        System.Reflection.ConstructorInfo? ctor = providerClass.GetConstructor(new[] { typeof(IHttpClientFactory) });
+        if (ctor != null)
+        {
+            BaseProvider? provider = (BaseProvider)ctor.Invoke(new object?[] { HttpClientFactory });
+            return provider;
```
gfs, line 60: Since we are caching the results of reflection shouldn't we cache this as well?

src/Shared/PackageManagers/PackageManagerFactory.cs
gfs: The only reference in the lib I can find to this is from DetectCryptographyTool. I too am not quite sure why you would want this interface instead of the one based on purl.

src/Shared/PackageManagers/PackageManagerFactory.cs
```diff
@@ -28,50 +30,39 @@ public static BaseProjectManager CreateBaseProjectManager(IHttpClientFactory htt
...
                 projectManagers.AddRange(typeof(BaseProjectManager).Assembly.GetTypes()
                .Where(type => type.IsSubclassOf(typeof(BaseProjectManager))));
             }
+
             // Use reflection to find the correct package management class
-            Type? downloaderClass = projectManagers
-               .Where(type => type.Name.Equals($"{purl.Type}ProjectManager",
-                                               StringComparison.InvariantCultureIgnoreCase))
-               .FirstOrDefault();
+            Type? managerClass = projectManagers
```
gfs, line 52: This reflection is potentially run many times.

gfs (changes requested):
See comments throughout and agree with the other comments left as well.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -113,8 +114,8 @@ public static class MutateExtension
                 foreach (Mutation mutation in mutator.Generate(nameToMutate))
                 {
                     // Construct the mutated name if the package was scoped.
-                    string? mutationNamespace = isScoped ? mutation.Mutated : null;
-                    string mutationName = isScoped ? purl.Name : mutation.Mutated;
+                    string? mutationNamespace = HttpUtility.UrlEncode(isScoped ? mutation.Mutated : null);
```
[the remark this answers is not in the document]
jpinz: I had a case where `newtonsoft.json` was mutated into `newtonsoft/json`, which when parsed back into a package url became `pkg:nuget/newtonsoft/json` which would be interpreted as the namespace being `newtonsoft`, and the name being `json`. We might want to change that on the mutator level to be impossible to do, but this also fixes it.

src/Shared/Model/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+    public BaseProvider() : this(new DefaultHttpClientFactory())
+    {}
+
+    /// <inheritdoc />
+    public IHttpClientFactory HttpClientFactory { get; protected init; }
+
+    /// <inheritdoc />
+    public virtual Task<string?> DownloadAsync(BaseProjectManager projectManager, PackageURL packageUrl, string targetDirectory, bool doExtract,
+        bool cached = false, CancellationToken cancellationToken = default) =>
+        throw new System.NotImplementedException();
```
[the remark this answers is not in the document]
jpinz: Good call!

src/Shared/Model/Providers/NuGetProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+public class NuGetProvider : BaseProvider
+{
+    private readonly SourceCacheContext _sourceCacheContext = new();
+    private readonly SourceRepository _sourceRepository = Repository.Factory.GetCoreV3("https://api.nuget.org/v3/index.json");
+    private readonly NuGetLogger _logger = new(LogManager.GetCurrentClassLogger());
+
+    /// <summary>
+    /// Instantiates a new <see cref="NuGetProvider"/>.
+    /// </summary>
+    public NuGetProvider(IHttpClientFactory? httpClientFactory = null)
```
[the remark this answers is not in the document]
jpinz: done!

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: It isn't. You need a PackageUrl with a name and version specified in order to download said package version.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: purl shouldn't ever be null. I did just change it though. Now it is `if (string.IsNullOrWhiteSpace(purl.Name))`

src/Shared/PackageManagers/NuGetProjectManager.cs
gfs: I think the point was that the purl itself is coming from a public api - even though you have declared it to not be null in the interface you can't trust that a user won't give you a null.

To be fair though it would actually crash on line 95 when you print the name wihout the null guard.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -84,16 +83,19 @@ private async Task<string> GetRegistrationEndpointAsync()
...
-        /// <returns> n/a </returns>
+        /// <remarks>The target directory is defined when creating the <see cref="NuGetProjectManager"/> in the subdirectory named `nuget-{packagename}@{packageversion}`</remarks>
+        /// <param name="purl">The <see cref="PackageURL"/> of the package to download.</param>
+        /// <param name="doExtract">If the contents of the .nupkg should be extracted into a directory.</param>
+        /// <param name="cached">If the downloaded contents should be retrieved from the cache if they exist there.</param>
+        /// <returns>An <see cref="IEnumerable{String}"/> list of the path(s) the contents were downloaded to.</returns>
         public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl, bool doExtract, bool cached = false)
         {
-            Logger.Trace("DownloadVersion {0}", purl?.ToString());
+            Logger.Trace("DownloadVersion {0}", purl.ToString());
```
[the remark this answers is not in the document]
gfs: In C#11 you will be able to do this instead with:
```
public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl!!, bool doExtract, bool cached = false)
```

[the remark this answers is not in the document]
gfs: Actually, we could enable `!!` by adding `<EnablePreviewFeatures>True</EnablePreviewFeatures>` to the csproj.

https://dotnetcoretutorials.com/2022/02/18/c11-parameter-null-checking/

src/Shared/Model/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using Contracts;
+using PackageManagers;
+using PackageUrl;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+public class BaseProvider : IManagerProvider<IManagerMetadata>
```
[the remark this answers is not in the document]
jpinz: Class was removed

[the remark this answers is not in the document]
jpinz: Class removed. NoOpProvider covers what I was using this for.

src/Shared/Model/Providers/NuGetProvider.cs
jpinz: Made it a static method now. Still in BaseProjectManager though. Maybe I should make an ArchiveHelper class? Thoughts?

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -552,7 +556,7 @@ public virtual List<Version> GetVersions(JsonDocument? metadata)
 
         protected HttpClient CreateHttpClient()
         {
-            return HttpClientFactory.CreateClient(GetType().Name);
+            return Provider.CreateHttpClient(GetType().Name);
```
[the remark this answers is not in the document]
jpinz: Yeah.

src/Shared/PackageManagers/PackageManagerFactory.cs
jpinz: Removed

src/Shared/Contracts/IManagerProvider.cs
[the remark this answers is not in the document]
jpinz: Removed HttpClientFactory from the IManagerProvider.

src/Shared/Model/Providers/ProviderFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+            ManagerProviders.AddRange(typeof(BaseProvider).Assembly.GetTypes()
+                .Where(type => type.IsSubclassOf(typeof(BaseProvider))));
+        }
+
+        // Use reflection to find the correct provider class
+        Type? providerClass = ManagerProviders
+            .FirstOrDefault(type => type.Name.Equals($"{purl.Type}Provider",
+                StringComparison.InvariantCultureIgnoreCase));
+
+        if (providerClass == null)
```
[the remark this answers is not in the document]
jpinz: We now throw a new NotSupportedException(purl.Type) in this case.

src/Shared/Model/Providers/NuGetProvider.cs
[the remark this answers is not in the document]
jpinz: NuGetProvider doesn't take in an IHttpClientFactory anymore.

src/Shared/PackageManagers/BaseProjectManager.cs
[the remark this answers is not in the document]
jpinz: Removed HttpClientFactory from providers.

src/Shared/PackageManagers/NuGetProjectManager.cs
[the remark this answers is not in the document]
jpinz: I like that, will probably do that in a separate PR.

gfs:
Let me know when you'd like me to review again.

src/Shared/PackageManagers/PackageManagerFactory.cs
```diff
@@ -28,50 +30,39 @@ public static BaseProjectManager CreateBaseProjectManager(IHttpClientFactory htt
...
         /// Get an appropriate project manager for package given its PackageURL.
         /// </summary>
         /// <param name="purl">The <see cref="PackageURL"/> for the package to create the project manager for.</param>
-        /// <param name="httpClientFactory"> The <see cref="IHttpClientFactory"/> for the project manager to use for making Http Clients to make web requests.</param>
+        /// <param name="managerProviderFactory"> The <see cref="IManagerProviderFactory"/> for the project manager to use for making the package manager's provider.</param>
         /// <param name="destinationDirectory">The directory to use to store any downloaded packages.</param>
-        /// <returns> BaseProjectManager object </returns>
-        public static BaseProjectManager? CreateProjectManager(PackageURL purl, IHttpClientFactory? httpClientFactory = null, string? destinationDirectory = null)
+        /// <returns>A new <see cref="BaseProjectManager"/> object implementation.</returns>
+        public static BaseProjectManager? CreateProjectManager(PackageURL purl, IManagerProviderFactory? managerProviderFactory = null, string? destinationDirectory = null)
```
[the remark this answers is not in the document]
jpinz: Renamed to ProjectManagerFactory.
The IManagerProviderFactory is now gone, I refactored it all again.

gfs:
Changes are major enough that I will need to pull down and poke around in my IDE.  Will review tomorrow.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+        private delegate BaseProjectManager? ConstructProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory = ".", IManagerProvider? managerProvider = null);
+
+        private readonly IHttpClientFactory _httpClientFactory;
+        private readonly IManagerProvider? _nuGetProvider;
+        
+        /// <summary>
+        /// The dictionary of project managers.
+        /// </summary>
+        private static readonly Dictionary<string, ConstructProjectManager> ProjectManagers =
+            new(StringComparer.InvariantCultureIgnoreCase)
```
pmalmsten, line 23: As we discussed offline, if this field were non-static and initialized in the constructor instead, the lambdas here would be able to directly access the singleton services they need instead of needing to pass them in via the `ConstructProjectManager` delegate.

By doing that, we could drop the two private fields (for `IHttpClientFactory` and `IManagerProvider?`, drop the corresponding two arguments from `ConstructProjectManager`, etc.

src/Shared/Contracts/IManagerProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+/// <summary>
+/// The interface to implement project manager specific methods.
+/// Useful for mocking when doing unit tests.
+/// </summary>
+public interface IManagerProvider
```
pmalmsten, line 15: Nit: the name `IManagerProvider` sounds to me like something that can provide Managers (e.g. a manager factory). That doesn't really seem like what implementations of this type is doing - it seems like the methods here are more focused on particular package actions? 

I wonder if there could be a better name that conveys that more clearly - as a brainstorm, would something like `IManagerPackageActions` make sense?

src/Shared/Model/Metadata/NuGetProtocolMetadata.cs
```diff
@@ -0,0 +1,145 @@
...
+using NuGet.Protocol;
+using NuGet.Versioning;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+/// <summary>
+/// A class to represent Package Metadata for a NuGet package.
+/// </summary>
+public record NuGetMetadata : IManagerMetadata
```
pmalmsten, line 18: Nit: should we emphasize the package-version-specific nature of this metadata? Maybe call it something like `NuGetPackageVersionMetadata`?

src/Shared/Model/Providers/NuGetProvider.cs
```diff
@@ -0,0 +1,162 @@
...
+    private readonly SourceCacheContext _sourceCacheContext = new();
+    private readonly SourceRepository _sourceRepository = Repository.Factory.GetCoreV3("https://api.nuget.org/v3/index.json");
+    private readonly NuGetLogger _logger = new(LogManager.GetCurrentClassLogger());
+
+    /// <summary>
+    /// Instantiates a new <see cref="NuGetProvider"/>.
+    /// </summary>
+    public NuGetProvider()
+    {
+    }
```
pmalmsten, line 33: Nit: this looks equivalent to the default constructor that C# automatically provides. Should we delete this in favor of the default constructor?

src/Shared/Model/Providers/NuGetProvider.cs
```diff
@@ -0,0 +1,162 @@
...
+        string topLevelDirectory,
+        string targetDirectory,
+        bool doExtract,
+        bool cached = false,
+        CancellationToken cancellationToken = default)
+    {
+        FindPackageByIdResource resource = await _sourceRepository.GetResourceAsync<FindPackageByIdResource>();
+
+        // Create a new memory stream to populate with the .nupkg.
+        using MemoryStream packageStream = new();
```
pmalmsten, line 47: How large can NuGet package files get? Have we considered something like downloading to a temporary file instead of writing to RAM?

pmalmsten: Looks like there are some decent ways to go about using a temporary file: https://stackoverflow.com/questions/34776031/how-to-create-a-temp-file-that-is-automatically-deleted-after-the-program-termin

pmalmsten: This point could easily be addressed in a separate PR though - not a blocker here.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -2,39 +2,42 @@
...
         {
+            ManagerProvider = provider ?? new NuGetProvider();
             GetRegistrationEndpointAsync().Wait();
         }
 
         public NuGetProjectManager(string destinationDirectory) : base(destinationDirectory)
         {
+            ManagerProvider = new NuGetProvider();
             GetRegistrationEndpointAsync().Wait();
         }
```
pmalmsten, line 42: Nit: Could we simply make the `IManagerProvider` required? It seems like the only existing uses always supply the argument.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -2,39 +2,42 @@
...
         public const string ENV_NUGET_ENDPOINT_API = "https://api.nuget.org";
         public const string ENV_NUGET_ENDPOINT = "https://www.nuget.org";
         public const string NUGET_DEFAULT_REGISTRATION_ENDPOINT = "https://api.nuget.org/v3/registration5-gz-semver2/";
+        private const string NUGET_DEFAULT_CONTENT_ENDPOINT = "https://api.nuget.org/v3-flatcontainer/";
 
         private string? RegistrationEndpoint { get; set; } = null;
+        private IManagerProvider ManagerProvider { get; }
 
-        public NuGetProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory) : base(httpClientFactory, destinationDirectory)
+        public NuGetProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory, IManagerProvider? provider = null) : base(httpClientFactory, destinationDirectory)
```
pmalmsten, line 32: Will this class function properly if any other implementation of `IManagerProvider` is provided? If not, should we use a narrower interface type that only `NuGetProvider` implements? 

(The fundamental idea of interfaces is to allow for interchanging of different implementations that all satisfy the same contract - if this class would not work if (say) a future `NpmProvider` was provided here, then we should have a narrower interface type specific just to NuGet that this class accepts instead. Such a narrower interface type could extend from `IManagerProvider` if we want to share common method signatures).

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -292,39 +178,31 @@ public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL pur
...
+                    packageVersion = latestVersion;
+                }
 
-                string? content = await GetHttpStringCache(httpClient, versionUri, useCache);
-                if (string.IsNullOrEmpty(content)) { return null; }
+                // Construct a new PackageURL that's guaranteed to have a  version.
+                PackageURL purlWithVersion = new (purl.Type, purl.Namespace, packageName, packageVersion, purl.Qualifiers, purl.Subpath);
+                
+                NuGetMetadata? packageVersionMetadata =
+                    await ManagerProvider.GetMetadataAsync(purlWithVersion, useCache: useCache) as NuGetMetadata;
```
pmalmsten, line 199: Regarding casting the result to `NuGetMetadata` - this goes with my earlier comment about other implementations of `IManagerProvider` not necessarily being suitable for use with this class. If we had a Nuget-specific version of `IManagerProvider`, that would eliminate the need for casting the return value here, since the Nuget-specific `IManagerProvider` would just return `NuGetMetadata` directly.

pmalmsten (changes requested):

src/Shared/Contracts/IManagerMetadata.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+
+namespace Microsoft.CST.OpenSource.Contracts;
+
+public interface IManagerMetadata
```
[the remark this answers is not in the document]
jpinz: I could add some documentation for it. But the idea here is this is the interface that ProjectManagers will implement for whatever metadata is returned by the registry API. It won't contain all of the information we want, just whatever the registry returns. So for example NuGet's current implementation doesn't contain a URI to download the nupkg, or any digest for the package, but we want that data. It just doesn't get returned from the API we are using.
So I think the interface name is appropriate, but I will add some docs to explain it better.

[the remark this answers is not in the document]
jpinz: Here's what I have right now:
"An interface representing the metadata returned from a project manager's API for metadata, won't contain extra or calculated properties."

src/Shared/Contracts/IManagerProvider.cs
jpinz: @brbayes-msft and I were chatting about this. We didn't have a better name at the time. I like your suggestion.
Any thoughts @gfs ?

src/Shared/Model/Metadata/NuGetProtocolMetadata.cs
jpinz: I renamed it, and I ended up renaming the interface to `IManagerPackageVersionMetadata`

src/Shared/Contracts/IManagerMetadata.cs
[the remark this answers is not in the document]
jpinz: I changed my mind and renamed it to `IManagerPackageVersionMetadata` it's a bit wordy, but makes more sense.

src/Shared/Contracts/IManagerProvider.cs
jpinz: I renamed it to `IManagerPackageActions`, we can change it pretty easily if we decide otherwise. But it makes a lot more sense than `IManagerProvider`

src/Shared/Model/Providers/NuGetProvider.cs
jpinz: Probably will address in another PR. But this implementation is straight from the documentation just as an FYI:
https://docs.microsoft.com/en-us/nuget/reference/nuget-client-sdk#download-a-package

src/Shared/Contracts/IManagerMetadata.cs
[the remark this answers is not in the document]
pmalmsten: `IManagerPackageVersionMetadata` sounds pretty decent to me.

src/Shared/Model/Metadata/NuGetPackageVersionPackageVersionMetadata.cs
```diff
@@ -13,9 +13,9 @@ namespace Microsoft.CST.OpenSource.Model.Metadata;
 using System.ComponentModel;
 
 /// <summary>
-/// A class to represent Package Metadata for a NuGet package.
+/// A class to represent Package Metadata for a NuGet package version.
 /// </summary>
-public record NuGetMetadata : IManagerMetadata
+public record NuGetPackageVersionPackageVersionMetadata : IManagerPackageVersionMetadata
```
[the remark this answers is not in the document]
pmalmsten: Looks like this was fixed in the most recent update actually - resolving this thread.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -292,39 +178,31 @@ public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL pur
...
+                    packageVersion = latestVersion;
+                }
 
-                string? content = await GetHttpStringCache(httpClient, versionUri, useCache);
-                if (string.IsNullOrEmpty(content)) { return null; }
+                // Construct a new PackageURL that's guaranteed to have a  version.
+                PackageURL purlWithVersion = new (purl.Type, purl.Namespace, packageName, packageVersion, purl.Qualifiers, purl.Subpath);
+                
+                NuGetPackageVersionMetadata? packageVersionMetadata =
+                    await NuGetPackageActions.GetMetadataAsync(purlWithVersion, useCache: useCache) as NuGetPackageVersionMetadata;
```
pmalmsten, line 199: Nit: casting is no longer required now that the interface is generic.

Same note applies in a few places in this file.

pmalmsten (commented):
Recent changes look pretty good - just gave you one more nit.

src/Shared/PackageManagers/CPANProjectManager.cs
```diff
@@ -120,7 +120,7 @@ public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl,
                 }
                 if (doExtract)
                 {
-                    downloadedPaths.Add(await ExtractArchive(targetName, await result.Content.ReadAsByteArrayAsync(), cached));
+                    downloadedPaths.Add(await ExtractArchive(TopLevelExtractionDirectory, targetName, await result.Content.ReadAsByteArrayAsync(), cached));
```
gfs, line 123: Why are we converting the Stream to an array? I think this work if you just pass the stream instead of reading hteb ytes and then making a new stream to hold it.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -323,7 +324,7 @@ public async Task<string> ExtractArchive(string directoryName, byte[] bytes, boo
                 Parallel = true
                 // MaxExtractedBytes = 1000 * 1000 * 10;  // 10 MB maximum package size
             };
-            ExtractionStatusCode result = await extractor.ExtractToDirectoryAsync(TopLevelExtractionDirectory, dirBuilder.ToString(), new MemoryStream(bytes), extractorOptions);
+            ExtractionStatusCode result = await extractor.ExtractToDirectoryAsync(topLevelDirectory, dirBuilder.ToString(), new MemoryStream(bytes), extractorOptions);
```
gfs, line 327: You end up turning the bytes back into a Stream before calling extract anyway. Recommend you just pass the Stream.

src/Shared/PackageManagers/CPANProjectManager.cs
gfs: If that has not been working with RecursiveExtractor let me know and I can investigate.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -2,39 +2,42 @@
...
         public const string NUGET_DEFAULT_REGISTRATION_ENDPOINT = "https://api.nuget.org/v3/registration5-gz-semver2/";
+        private const string NUGET_DEFAULT_CONTENT_ENDPOINT = "https://api.nuget.org/v3-flatcontainer/";
 
         private string? RegistrationEndpoint { get; set; } = null;
+        private NuGetPackageActions NuGetPackageActions { get; }
 
-        public NuGetProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory) : base(httpClientFactory, destinationDirectory)
+        public NuGetProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory, NuGetPackageActions nugetPackageActions) : base(httpClientFactory, destinationDirectory)
         {
+            NuGetPackageActions = nugetPackageActions;
```
gfs, line 34: Class name and Instance variable of that class are named identically.  This is a private (also probably readonly?) so naming pattern would normally be `_nuGetPackageActions`

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -337,267 +215,199 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
+            PackageURL purlWithVersion = !string.IsNullOrWhiteSpace(purl.Version) ? 
+                purl : new PackageURL(purl.Type, purl.Namespace, purl.Name, latestVersion, purl.Qualifiers, purl.Subpath);
 
-            metadata.Name = latestCatalogEntry.GetProperty("id").GetString();
-            metadata.Description = latestCatalogEntry.GetProperty("description").GetString();
-            
-            // Title is different than name or description for NuGet packages, not always used.
-            // metadata.Title = GetLatestCatalogEntry(root).GetProperty("title").GetString();
+            NuGetPackageVersionMetadata packageVersionPackageVersionMetadata =
+                await NuGetPackageActions.GetMetadataAsync(purlWithVersion, useCache: useCache) as NuGetPackageVersionMetadata ?? throw new InvalidOperationException($"The metadata for {purlWithVersion} was null.");
```
gfs, line 225: Recommend returning `null` instead of throwing here.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// The dictionary of project managers.
+        /// </summary>
+        private readonly Dictionary<string, ConstructProjectManager> _projectManagers;
+
+        /// <summary>
+        /// Initializes a new instance of a <see cref="ProjectManagerFactory"/>.
+        /// </summary>
+        /// <param name="httpClientFactoryParam">The <see cref="IHttpClientFactory"/> to use in the project managers.</param>
+        /// <param name="nugetPackageActions">The <see cref="NuGetPackageActions"/> to use in the <see cref="NuGetProjectManager"/>.</param>
+        public ProjectManagerFactory(IHttpClientFactory? httpClientFactoryParam = null, NuGetPackageActions? nugetPackageActions = null)
```
gfs, line 30:
```suggestion
        public ProjectManagerFactory(IHttpClientFactory? httpClientFactoryParam = null, IManagerPackageActions? managerPackageActions = null)
```

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -0,0 +1,128 @@
...
+                    nameof(MavenProjectManager), destinationDirectory =>
+                        new MavenProjectManager(httpClientFactory, destinationDirectory)
+                },
+                {
+                    nameof(NPMProjectManager), destinationDirectory =>
+                        new NPMProjectManager(httpClientFactory, destinationDirectory)
+                },
+                {
+                    nameof(NuGetProjectManager), destinationDirectory =>
+                        new NuGetProjectManager(httpClientFactory, destinationDirectory, nugetPackageActions ?? new NuGetPackageActions()) // Add the NuGetPackageActions to the NuGetProjectManager.
```
gfs, line 82:
```suggestion
                        new NuGetProjectManager(httpClientFactory, destinationDirectory, managerPackageActions is null ? new NugetPackageActions() : managerPackageActions is NugetPackageActions npa ? npa : throw new ArgumentException(nameof(managerPackageActions)) // Add the NuGetPackageActions to the NuGetProjectManager.
```

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /// <summary>
+        /// Add static method to just get a <see cref="BaseProjectManager"/> implementation.
+        /// </summary>
+        /// <param name="packageUrl">The <see cref="PackageURL"/> to get a project manager for.</param>
+        /// <param name="httpClientFactory">The <see cref="IHttpClientFactory"/> to optionally add.</param>
+        /// <param name="nugetPackageActions">The <see cref="NuGetPackageActions"/> to use in the <see cref="NuGetProjectManager"/>.</param>
+        /// <returns>A new <see cref="BaseProjectManager"/> implementation.</returns>
+        public static BaseProjectManager? GetProjectManager(PackageURL packageUrl, IHttpClientFactory? httpClientFactory = null, NuGetPackageActions? nugetPackageActions = null)
```
gfs, line 123:
```suggestion
        public static BaseProjectManager? GetProjectManager(PackageURL packageUrl, IHttpClientFactory? httpClientFactory = null, IManagerPackageActions? managerPackageActions = null)
```

src/oss-characteristics/CharacteristicTool.cs
```diff
@@ -73,11 +74,15 @@ public static IEnumerable<Example> Examples
             public FailureLevel SarifLevel { get; set; } = FailureLevel.Note;
         }
 
+        public CharacteristicTool(ProjectManagerFactory projectManagerFactory) : base(projectManagerFactory)
+        {
+        }
+
         public CharacteristicTool(IHttpClientFactory httpClientFactory) : base(httpClientFactory)
```
gfs, line 81: Remove

gfs: I think we should be removing all the IHttpClient based ctors.

gfs: Or like below it should make a projectmanagerfactory with that httpclient to maintain compat.

gfs, line 81:
```suggestion
        public CharacteristicTool(IHttpClientFactory httpClientFactory) : base(new ProjectManagerFactory(httpClientFactory))
```

src/oss-find-source/FindSourceTool.cs
```diff
@@ -23,11 +24,15 @@ public class FindSourceTool : OSSGadget
             "pkg:github/metacpan/metacpan-web"
         };
 
+        public FindSourceTool(ProjectManagerFactory projectManagerFactory) : base(projectManagerFactory)
+        {
+        }
+
         public FindSourceTool(IHttpClientFactory httpClientFactory) : base(httpClientFactory)
```
gfs, line 31:
```suggestion
        public FindSourceTool(IHttpClientFactory httpClientFactory) : this(new ProjectManagerFactory(httpClientFactory))
```

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.CST.OpenSource.FindSquats.ExtensionMethods
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading;
+    using System.Web;
```
gfs, line 14: ? Dependency change with no code change?

src/oss-health/HealthTool.cs
```diff
@@ -17,22 +17,28 @@ namespace Microsoft.CST.OpenSource
 
     public class HealthTool : OSSGadget
     {
+        
+        public HealthTool(ProjectManagerFactory projectManagerFactory) : base(projectManagerFactory)
+        {
+        }
+
         public HealthTool(IHttpClientFactory httpClientFactory) : base(httpClientFactory)
```
gfs, line 25:
```suggestion
        public HealthTool(IHttpClientFactory httpClientFactory) : this(new ProjectManagerFactory(httpClientFactory))
```

gfs: Unless there is something I'm missing this seems like a better pattern.  At the very least we need to choose one to use because there are both ones that case base with httpclientfactory and those that make a projectmanagerfactory based on the httpclient.

src/oss-reproducible/ReproducibleTool.cs
```diff
@@ -75,12 +75,18 @@ public static IEnumerable<Example> Examples
             public bool LeaveIntermediateFiles { get; set; }
         }
 
+        public ReproducibleTool(ProjectManagerFactory projectManagerFactory) : base(projectManagerFactory)
+        {
+        }
+
         public ReproducibleTool(IHttpClientFactory httpClientFactory) : base(httpClientFactory)
```
gfs, line 82: +

src/oss-risk-calculator/RiskCalculatorTool.cs
```diff
@@ -67,11 +68,15 @@ public static IEnumerable<Example> Examples
             public bool UseCache { get; set; }
         }
 
+        public RiskCalculatorTool(ProjectManagerFactory projectManagerFactory) : base(projectManagerFactory)
+        {
+        }
+
         public RiskCalculatorTool(IHttpClientFactory httpClientFactory) : base(httpClientFactory)
```
gfs, line 75: +

src/oss-tests/FindSquatsTests.cs
```diff
@@ -23,7 +27,7 @@ namespace Microsoft.CST.OpenSource.Tests
     public class FindSquatsTest
     {
         [DataTestMethod]
-        [DataRow("pkg:nuget/Microsoft.CST.OAT", false)]
+        // [DataRow("pkg:nuget/Microsoft.CST.OAT", false)] TODO: Figure out why this fails on ADO.
```
gfs, line 30: Does this only fail in pipeline?

gfs (changes requested):

src/oss-tests/FindSquatsTests.cs
jpinz: Yes. I have yet to be able to repro it locally.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -0,0 +1,139 @@
...
+                        new URLProjectManager(httpClientFactory, destinationDirectory)
+                },
+                {
+                    nameof(VSMProjectManager),
+                    (httpClientFactory, destinationDirectory, _) =>
+                        new VSMProjectManager(httpClientFactory, destinationDirectory)
+                },
+            };
+
+        public ProjectManagerFactory(IHttpClientFactory? httpClientFactory = null, IManagerProvider? nuGetProvider = null)
```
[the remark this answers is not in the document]
jpinz: ManagerProvider is no longer the naming scheme, it's now PackageActions. Did you review an old version of the code?

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: The issue with the factory now is that it's not exactly DI, but trying to mimic how DI would work. So in that case, we would need one parameter for each project manager that uses a `IManagerPackageActions` in the ProjectManager implementation. So say one gets added for NPM, the `ProjectManagerFactory` constructor would have to be:
`public ProjectManagerFactory(IHttpClientFactory? httpClientFactoryParam = null, NuGetPackageActions? nugetPackageActions = null, NpmPackageActions? npmPackageActions = null)` so that way we can override one, the other, both, or neither.

jpinz: In other words, I don't think we can change it to just `managerPackageActions`, as it will have to be manager specific at least given this new design.

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: I don't think this is applicable based on the comment I left: https://github.com/microsoft/OSSGadget/pull/318#discussion_r841916574

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: Also don't think this is applicable base on my previous comment: https://github.com/microsoft/OSSGadget/pull/318#discussion_r841916574

src/oss-health/HealthTool.cs
jpinz: Yeah I switched all of the OSSGadget CLI tools to this pattern.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -289,15 +289,16 @@ public virtual Task<IEnumerable<string>> EnumerateVersions(PackageURL purl, bool
...
         /// recursively, using RecursiveExtractor.
         /// </summary>
-        /// <param name="directoryName">directory to extract content into (within TopLevelExtractionDirectory)</param>
+        /// <param name="topLevelDirectory">The top level directory content should be extracted to.</param>
+        /// <param name="directoryName">directory to extract content into (within <paramref name="topLevelDirectory"/>)</param>
         /// <param name="bytes">bytes to extract (should be an archive file)</param>
         /// <param name="cached">If the archive has been cached.</param>
         /// <returns></returns>
-        public async Task<string> ExtractArchive(string directoryName, byte[] bytes, bool cached = false)
+        public static async Task<string> ExtractArchive(string topLevelDirectory, string directoryName, byte[] bytes, bool cached = false)
```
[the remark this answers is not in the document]
jpinz: Refactored to an ArchiveHelper.

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: Switched to passing the stream around instead.

src/Shared.CLI/OSSGadget.cs
```diff
@@ -21,6 +21,10 @@ public class OSSGadget : OssGadgetLib
         public static string ToolName { get => GetToolName() ?? ""; }
         public static string ToolVersion { get => GetToolVersion() ?? ""; }
 
+        public OSSGadget(ProjectManagerFactory projectManagerFactory) : base(projectManagerFactory)
+        {
+        }
+
         public OSSGadget(IHttpClientFactory httpClientFactory) : base(httpClientFactory)
```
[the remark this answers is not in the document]
jpinz: Remove the ProjectManagerFactory one? or the IHttpClientFactory one?
I removed the IHttpClientFactory one based on some of your later comments.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -2,39 +2,42 @@
...
         public const string ENV_NUGET_ENDPOINT_API = "https://api.nuget.org";
         public const string ENV_NUGET_ENDPOINT = "https://www.nuget.org";
         public const string NUGET_DEFAULT_REGISTRATION_ENDPOINT = "https://api.nuget.org/v3/registration5-gz-semver2/";
+        private const string NUGET_DEFAULT_CONTENT_ENDPOINT = "https://api.nuget.org/v3-flatcontainer/";
 
         private string? RegistrationEndpoint { get; set; } = null;
+        private NuGetPackageActions _nuGetPackageActions { get; }
 
-        public NuGetProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory) : base(httpClientFactory, destinationDirectory)
+        public NuGetProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory, NuGetPackageActions nugetPackageActions) : base(httpClientFactory, destinationDirectory)
```
gfs, line 32:
```suggestion
        public NuGetProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory, IManagerPackageActions<NuGetPackageVersionMetadata> nugetPackageActions) : base(httpClientFactory, destinationDirectory)
```

gfs: As long as something implements the manager actions that return nuget metadata we don't need to require it be the specific `NuGetPackageActions` implementation do we?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -208,77 +145,26 @@ public override async Task<bool> PackageExists(PackageURL purl, bool useCache =
...
         /// <inheritdoc />
         public override async Task<IEnumerable<string>> EnumerateVersions(PackageURL purl, bool useCache = true)
         {
-            Logger.Trace("EnumerateVersions {0}", purl?.ToString());
+            Logger.Trace("EnumerateVersions {0}", purl.ToString());
 
-            if (purl == null || purl.Name is null)
+            if (string.IsNullOrWhiteSpace(purl.Name))
             {
                 return new List<string>();
```
gfs, line 159:
```suggestion
                return Array.Empty<string>();
```
Less allocation overhead than providing a list.

src/oss-detect-cryptography/DetectCryptographyTool.cs
```diff
@@ -92,9 +93,8 @@ private static async Task Main(string[] args)
                             {
                                 targetDirectoryName = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
                             }
-                            BaseProjectManager projectManager = ProjectManagerFactory.CreateBaseProjectManager(detectCryptographyTool.HttpClientFactory, targetDirectoryName);
 
-                            string? path = await projectManager.ExtractArchive("temp", File.ReadAllBytes(target));
+                            string? path = await ArchiveHelper.ExtractArchiveAsync(targetDirectoryName, "temp", File.OpenRead(target));
```
gfs, line 97: Nit: As part of the underlying implementation in RecursiveExtractor your provided stream will be used copied to a new Stream inside a FileEntry, because we can't make assumptions about the incoming stream - like that it can be seeked - but doing the extraction requires those properties.

I thought that the Filename based api avoid this extra allocation but it looks like it does not so I have a PR to improve it: https://github.com/microsoft/RecursiveExtractor/pull/77.

You can either call RecursiveExtractor's ExtractToDirectory with a filename or with FileEntry.FromFileName

src/oss-diff/DiffTool.cs
```diff
@@ -85,11 +85,15 @@ public static IEnumerable<Example> Examples
             public IEnumerable<string> Targets { get; set; } = Array.Empty<string>();
         }
 
-        public DiffTool(IHttpClientFactory httpClientFactory) : base(httpClientFactory)
+        public DiffTool(ProjectManagerFactory projectManagerFactory) : base(projectManagerFactory)
         {
         }
 
-        public DiffTool() : this (new DefaultHttpClientFactory()) { }
+        public DiffTool(IHttpClientFactory httpClientFactory) : this(new ProjectManagerFactory(httpClientFactory))
```
gfs, line 92: I think all these ctors that take IHttpClientFactory should be removed.

src/Shared/PackageManagers/ProjectManagerFactory.cs
gfs: Given that the ctor has these tightly coupled arguments I don't think this is a good choice to use for the public API for the reasons you mention. This simply does not scale.

Here's my proposal similar to before, to help avoid some of this reflection:

I think this class should have this ctor

```
public ProjectManagerFactory(Dictionary<string, ConstructProjectManager>? startingManagerSet = null)
{
   // Keep the existing default behavior if you don't provide any managers, otherwise use the set the user provided
    _projectManagers = startingManagerSet ?? GenerateDefaultSet();
...
}
```

```
public Dictionary<string, ConstructProjectManager> GenerateDefaultSet(IHttpClientFactory? factory = null)
{
// Existing code that generates the set below
}
```

It should have these methods that manipulate the dictionary
```
// Replace the manager for the specified purl type.
public bool SetManager(string purlType, ConstructProjectManager constructMethod)
// Replace the dictionary with the provided dictionary
public bool SetManagers(IDictionary<string, ConstructProjectManager> managersToReplace)
// Remove the manager for the specified purl type
public bool RemoveManager(string purlType)
// Remove all entries from the dictionary
public bool ClearManagers()
```

src/Shared/PackageManagers/ProjectManagerFactory.cs
gfs: See newer comment. This is not a scalable approach and must be changed.

gfs (changes requested):

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -2,35 +2,33 @@
...
     using PackageUrl;
 
-    public class BaseProjectManager
+    public abstract class BaseProjectManager
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="BaseProjectManager"/> class.
         /// </summary>
-        public BaseProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory)
+        public BaseProjectManager(IHttpClientFactory httpClientFactory, string destinationDirectory = ".")
```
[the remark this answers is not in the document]
jpinz: No, because only NuGet right now uses any implementation of `IManagerPackageActions`, so there isn't a need to have it as part of the base project manager as well I don't think.
I separated the `IHttpClientFactory` out of the `IManagerPackageActions` (previously `IManagerProvider`) as it wasn't necessary to do so and caused some confusing scenarios where every project manager had a `IManagerPackageActions` but only used it as an object containing an `IHttpClientFactory`, adding an unnecessary wrapper essentially.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: We don't, I changed it. Fixed the failing unit tests that way too! Thanks haha.

src/Shared/PackageManagers/BaseProjectManager.cs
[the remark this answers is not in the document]
gfs: > No, because only NuGet right now uses any implementation of `IManagerPackageActions`, so there isn't a need to have it as part of the base project manager as well I don't think.

If the purpose of the refactor was to have an interface for provider actions but you only use it for Nuget its not an interface.  Its just class with a complex type.

>  I separated the `IHttpClientFactory` out of the `IManagerPackageActions` (previously `IManagerProvider`) as it wasn't necessary to do so and caused some confusing scenarios where every project manager had a `IManagerPackageActions` but only used it as an object containing an `IHttpClientFactory`, adding an unnecessary wrapper essentially.

I think this is far more confusing.

src/Shared/PackageManagers/NuGetProjectManager.cs
gfs: It should be `IManagerProvider<NugetProviderMetadata>` I think.

gfs:
Some other misc issues.

1. Does NugetPackageActions belong in the model directory? It seems like an implementation to me.
2. Why does NuGetPackageMetaData not derive from PackageMetadata?

src/Shared/Model/PackageActions/NuGetPackageActions.cs
```diff
@@ -0,0 +1,156 @@
...
+        string topLevelDirectory,
+        string targetDirectory,
+        bool doExtract,
+        bool cached = false,
+        CancellationToken cancellationToken = default)
+    {
+        FindPackageByIdResource resource = await _sourceRepository.GetResourceAsync<FindPackageByIdResource>();
+
+        // Create a new memory stream to populate with the .nupkg.
+        using MemoryStream packageStream = new();
```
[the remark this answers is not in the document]
jpinz: Good idea! Done.

gfs:
I did a partial implementation of the cargo manager in the proposed style that embraces the ManagerProvider interface.

https://github.com/microsoft/OSSGadget/pull/323/commits/16d089f66de6cfd1cc10f09da2f6b364a17bd833

In Doing so I noticed some potential concerns with that interface:
It returns `async Task<string>` what is the behavior supposed to be if you need to download multiple files?
Because it only returns string the doextract flag doesn't make sense.

Proposed either:

1. It should return IEnumerable<string> and keep the doextract flag
2. It keep returning string and remove the doextract flag
3. It returns bool and the location it is downloaded to is known by the caller because they passed it in.
4. The comments are changed to clarify that it returns a the path to a directory containing the all the downloaded files and not the files themselves.

src/oss-detect-cryptography/DetectCryptographyTool.cs
jpinz: > Some other misc issues.
> 
> 1. Does NugetPackageActions belong in the model directory? It seems like an implementation to me.
> 2. Why does NuGetPackageMetaData not derive from PackageMetadata?

The first one makes sense to me, I can refactor that.
NuGetPackageMetadata doesn't implement PackageMetadata as PackageMetadata was written as a "Catch-All" Metadata model, so it has unnecessary things that don't pertain to NuGet. The NuGetPackageMetadata in this case also doesn't represent what we want as an entire view of a NuGet package's metadata, currently what it is doing is just mapping the `PackageSearchMetadataRegistration` model from `NuGet.Protocol` to a model we control and can expand upon. Such as adding digest and hash data, plus some other data that we as a team have been discussing that we'd like as part of a package's metadata.
We're still flushing out final details, we'll loop you and @scovetta in once we're done to present to you both.

jpinz:
> 

I don't believe I changed any logic from what is currently in the repo, but I can change it to reflect your points here.
Currently, it does 4, which I think is fine personally, but the documentation should definitely be better to reflect that.

src/Shared/PackageManagers/BaseProjectManager.cs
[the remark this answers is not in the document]
jpinz: I understand the confusion, I was struggling to decide how to handle it myself.
I personally think that the `IManagerPackageActions` shouldn't contain an `IHttpClientFactory`, at least not currently.
I think one thing that could be worth doing is a follow up PR that implements a new `IManagerPackageActions` class for each manager and then separate out the logic accordingly. I think then it would make sense to have the `IHttpClientFactory` as a part of `IManagerPackageActions` but we would also want it in the `ProjectManager` implementation as well I'd imagine.

gfs:
Got it. Lets leave it as 4 then.

src/Shared/Model/Providers/NuGetProvider.cs
jpinz: Addressed.

src/Shared/Model/Metadata/NuGetPackageVersionMetadata.cs
```diff
@@ -0,0 +1,145 @@
...
+    [JsonProperty(PropertyName = JsonProperties.SubjectId)]
+    public Uri CatalogUri { get; init; }
+
+    /// <summary>
+    /// Initialize an instance of <see cref="NuGetPackageVersionMetadata"/> using the <see cref="JsonConstructorAttribute"/>.
+    /// </summary>
+    /// <remarks>Necessary for unit test implementation of json serialization and deserialization.</remarks>
+    [JsonConstructor]
+#pragma warning disable CS8618
+    public NuGetPackageVersionMetadata()
```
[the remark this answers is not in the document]
jpinz: So based on experimenting, and there could be another way of doing this, I believe I have to have this empty constructor in order to deserialize json. I can make it private and initialize all values as `null!` or set them to all be nullable to avoid the warning. But without this constructor, the unit tests at the json deserialization step fail.

src/Shared/PackageManagers/CPANProjectManager.cs
jpinz: I think it works fine now.

gfs:
I finished the cargo manager implementation and refactored it into a TypedManager class that can be used by the actions and also refactored the Nuget manager.  Take a look at the TypedManager.
'
In this manner each project manager does need much additional code and the base functionality is tightly coupled to the provider interface.

What do you think?

gfs:
Here's an updated compare of whats changed: https://github.com/microsoft/OSSGadget/pull/323/files/9036d201ac002649d35e944f457930ebc90d3c87..3b2dd5e615044b4ec8e594ace89d599b2b086d72

gfs:
image

I made a little psudocode snippet for my proposal for the factory.

src/Shared/Model/PackageActions/NuGetPackageActions.cs
```diff
@@ -0,0 +1,155 @@
...
+            _logger, 
+            cancellationToken);
+
+        return versions.Select(v => v.ToString());
+    }
+    
+    /// <inheritdoc />
+    public async Task<string> GetLatestVersionAsync(
+        PackageURL packageUrl,
+        bool includePrerelease = false,
```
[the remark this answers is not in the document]
jpinz: Implemented. and added to GetAllVersionsAsync as well and implemented as well.

jpinz:
> I finished the cargo manager implementation and refactored it into a TypedManager class that can be used by the actions and also refactored the Nuget manager. Take a look at the TypedManager. ' In this manner each project manager does need much additional code and the base functionality is tightly coupled to the provider interface.
> 
> What do you think?

I love this. I'll go mess around with it myself, but makes total sense to me and is much cleaner.

gfs:
> > I finished the cargo manager implementation and refactored it into a TypedManager class that can be used by the actions and also refactored the Nuget manager. Take a look at the TypedManager. ' In this manner each project manager does need much additional code and the base functionality is tightly coupled to the provider interface.
> > What do you think?
> 
> I love this. I'll go mess around with it myself, but makes total sense to me and is much cleaner.

I also did most of the refactor for the factory as proposed above in another commit, heres the updated compare. It should be keyed on purl.type rather than purl.typeProjectManager. The mechanism by which you get the values to put for the initial default is indecided. It could be a static value in the managers that are being set? We could create a type that has a parameter that matches the purl type? Or just hardcode the strings in the default method.

https://github.com/microsoft/OSSGadget/pull/323/files/9036d201ac002649d35e944f457930ebc90d3c87..8d5a3c5b52b535226bb789f1c7bbac6ba119d05a

jpinz:
Going to make a third PR after #326 goes in that covers the changes between that one and this one.

jpinz:
Going to make a third PR after #326 goes in that covers the changes between that one and this one.

PR #319, opened 2022-03-24
Add a new mutator to remove the namespace of a package.

jpinz:
This is in response to a new typo squatting attack done on @azure packages.
https://jfrog.com/blog/large-scale-npm-attack-targets-azure-developers-with-malicious-packages/


src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -104,32 +104,54 @@ public static class MutateExtension
...
+                    generatedMutations[newMutation.Mutated].Add(newMutation);
                 }
             }
 
+            if (!hasNamespace)
+            {
+                return generatedMutations;
+            }
+
+            PackageURL purlWithoutNamespace =
```
gfs, line 132: Why not make this its own mutate extension?

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -104,32 +104,54 @@ public static class MutateExtension
 
             Dictionary<string, IList<Mutation>> generatedMutations = new();
 
-            // Check to see if it is a scoped npm package to generate candidates for.
-            bool isScoped = purl.HasNamespace();
-            string nameToMutate = isScoped ? purl.Namespace : purl.Name;
+            // Check to see if the package has a namespace to generate candidates for.
+            bool hasNamespace = purl.HasNamespace();
```
gfs, line 108: This might be my fault based on the existing code, but this logic is actually very confusing and I think it could be improved just with an if(hasNamespace) splitting the two flows earlier.

src/oss-find-squats-lib/MutateExtension.cs
jpinz: I created a RemoveNamespaceMutator. It's kind of difficult to handle something like this because the design just mutates the namespace if there is one, so the mutators never get the full name (namespace+name). So I check after the if(!hasNamespace)

src/oss-find-squats-lib/MutateExtension.cs
jpinz: I tried to clean up the surrounding logic best I could. Let me know what you think.

gfs: Much cleaner.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -104,29 +104,31 @@ public static class MutateExtension
 
             Dictionary<string, IList<Mutation>> generatedMutations = new();
 
-            // Check to see if it is a scoped npm package to generate candidates for.
-            bool isScoped = purl.HasNamespace();
-            string nameToMutate = isScoped ? purl.Namespace : purl.Name;
+            // Check to see if the package has a namespace to generate candidates for.
```
gfs, line 107: Can you add a comment to this method that explicitly states the behavior.

i.e. if a Namespace is set mutations will only be generated based on mutations of the namespace, if not mutations will only be generated based on mutations of the package name.

src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -188,5 +190,23 @@ public static async IAsyncEnumerable<FindPackageSquatResult> EnumerateExistingSq
 
             }
         }
+
+        /// <summary>
+        /// Adds a <see cref="Mutation"/> to the dictionary of generated mutations.
+        /// </summary>
+        /// <param name="generatedMutations">The dictionary of generated mutations to add <paramref name="mutation"/> to.</param>
+        /// <param name="mutation">The <see cref="Mutation"/> to add to the dictionary.</param>
+        private static void AddMutation(this IDictionary<string, IList<Mutation>> generatedMutations, Mutation mutation)
```
gfs, line 199: nit: This might be a little cleaner with `AddOrUpdate` (requires ConcurrentDictionary) or `GetValueRefOrAdd`

for the second:
```c#
ref MyType value = CollectionsMarshal.TryGetValueRef(generatedMutations, mutationString, out bool exists);

if (exists)
{
    value.Add(mutation);
}
else
{
    ref value = new List<Mutation>() { mutation };
}

```



https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-6.0
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.getvaluereforadddefault?view=net-6.0
https://github.com/dotnet/runtime/issues/27062

src/oss-find-squats-lib/Mutators/RemoveNamespaceMutator.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    public class RemoveNamespaceMutator : IMutator
+    {
+        public MutatorType Kind { get; } = MutatorType.RemovedNamespace;
+
+        public IEnumerable<Mutation> Generate(string arg)
+        {
+            PackageURL purl = new(arg);
+            
+            yield return new Mutation(
```
gfs, line 20: Only should yield mutations if they are not the same as the original package. If you pass something with null namespace this should just yield break.

gfs (changes requested):
The new NamespaceRemovalMutator needs to yield break if provided a null namespace.

src/oss-find-squats-lib/MutateExtension.cs
jpinz: Managed to move this to the new Generate(PackageUrl) implementation, which I think is much cleaner.

gfs (approved):

gfs:
LGTM now.

PR #323, opened 2022-04-04
Gfs/cargo provider partial

gfs:
This is a POC implementation for using a `TypedManager` class to implement the default wrappers for the `IManagerActions` simplifying the Factory and Managers. Cargo and Nuget are implemented as examples.

Compare against #318, not main.


src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -45,7 +44,7 @@ public BaseProjectManager(string destinationDirectory = ".") : this(new DefaultH
         /// <summary>
         /// The <see cref="IHttpClientFactory"/> for the manager.
         /// </summary>
-        public IHttpClientFactory HttpClientFactory { get; }
+        public IHttpClientFactory HttpClientFactory { get; init; }
```
gfs, line 47: This change probably wasn't needed - ctors should be refactored so it isn't needed.

jpinz:
@gfs is this still needed?

gfs:
> @gfs is this still needed?

No.

PR #324, opened 2022-04-05
Refactor ExtractArchive into a helper class, and a few other changes.

jpinz:
A lot of this work comes from work I had done in #318 that I just wanted to separate out to make that PR easier to review.

This PR includes:

- Refactor `ExtractArchive` from `BaseProjectManager` into a separate `ArchiveHelper` static class.
- Rename `PackageManagerFactory` to `ProjectManagerFactory`, and remove the `BaseProjectManager` constructors as they were able to be removed thanks to the `ExtractArchive` change detailed above.
- Updated `PackageMetadata.User` to be a record instead of a class, will help with future json parsing.
- Made `BaseProjectManager` abstract.
- Add a `Type` readonly string to each implementation of `BaseProjectManager`. Isn't currently being used as of this PR, but #318 uses it, and I anticipate it being useful in other places in the future as well.
- Add `includePrerelease` to `BaseProjectManager.EnumerateVersions` to filter out pre-release/beta versions if they aren't wanted. Defaults to `true`, so there are no changes from what is currently in production.
- Make `BaseProjectManager.GetPackageMetadata` return a nullable `PackageMetadata` object.
- Fix any potential issues down the line in `IMutator` with mutated names including things that aren't HTTP encoded. 
   - I was running into an issue mutating `newtonsoft.json` as it was creating `newtonsoft/json` as a mutation because `.` is close to `/` and that was causing issues with `newtonsoft` becoming the namespace, and `json` becoming the name.

### DIFF
--- a/src/Shared/Helpers/ArchiveHelper.cs
+++ b/src/Shared/Helpers/ArchiveHelper.cs
@@ -24,7 +24,10 @@ public static class ArchiveHelper
     /// <param name="content">stream of the contents to extract (should be an archive file)</param>
     /// <param name="cached">If the archive has been cached.</param>
     /// <returns>The path that the archive was extracted to.</returns>
-    public static async Task<string> ExtractArchiveAsync(string topLevelDirectory, string directoryName, Stream content,
+    public static async Task<string> ExtractArchiveAsync(
+        string topLevelDirectory,
+        string directoryName,
+        Stream content,
         bool cached = false)
     {
         Logger.Trace("ExtractArchive({0}, <stream> len={1})", directoryName, content.Length);
--- a/src/Shared/Helpers/ArchiveHelper.cs
+++ b/src/Shared/Helpers/ArchiveHelper.cs
@@ -1,0 +1,71 @@
+// Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+
+namespace Microsoft.CST.OpenSource.Helpers;
+
+using RecursiveExtractor;
+using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+public static class ArchiveHelper
+{
+    /// <summary>
+    /// Logger for each of the subclasses
+    /// </summary>
+    static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();
+
+    /// <summary>
+    /// Extracts an archive (given by 'bytes') into a directory named 'directoryName',
+    /// recursively, using RecursiveExtractor.
+    /// </summary>
+    /// <param name="topLevelDirectory">The top level directory content should be extracted to.</param>
+    /// <param name="directoryName">directory to extract content into (within <paramref name="topLevelDirectory"/>)</param>
+    /// <param name="content">stream of the contents to extract (should be an archive file)</param>
+    /// <param name="cached">If the archive has been cached.</param>
+    /// <returns>The path that the archive was extracted to.</returns>
+    public static async Task<string> ExtractArchiveAsync(string topLevelDirectory, string directoryName, Stream content,
+        bool cached = false)
+    {
+        Logger.Trace("ExtractArchive({0}, <stream> len={1})", directoryName, content.Length);
+
+        Directory.CreateDirectory(topLevelDirectory);
+
+        StringBuilder dirBuilder = new(directoryName);
+
+        foreach (char c in Path.GetInvalidPathChars())
+        {
+            dirBuilder.Replace(c, '-'); // ignore: lgtm [cs/string-concatenation-in-loop]
+        }
+
+        string fullTargetPath = Path.Combine(topLevelDirectory, dirBuilder.ToString());
+
+        if (!cached)
+        {
+            while (Directory.Exists(fullTargetPath) || File.Exists(fullTargetPath))
+            {
+                dirBuilder.Append("-" + DateTime.Now.Ticks);
+                fullTargetPath = Path.Combine(topLevelDirectory, dirBuilder.ToString());
+            }
+        }
+
+        Extractor extractor = new();
+        ExtractorOptions extractorOptions = new()
+        {
+            ExtractSelfOnFail = true, Parallel = true
+            // MaxExtractedBytes = 1000 * 1000 * 10;  // 10 MB maximum package size
+        };
+        ExtractionStatusCode result = await extractor.ExtractToDirectoryAsync(topLevelDirectory, dirBuilder.ToString(),
+            content, extractorOptions);
+        if (result == ExtractionStatusCode.Ok)
+        {
+            Logger.Debug("Archive extracted to {0}", fullTargetPath);
+        }
+        else
+        {
+            Logger.Warn("Error extracting archive {0} ({1})", fullTargetPath, result);
+        }
+
+        return fullTargetPath;
+    }
+} 
--- a/src/Shared/PackageManagers/BaseProjectManager.cs
+++ b/src/Shared/PackageManagers/BaseProjectManager.cs
@@ -21,7 +21,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         /// The type of the project manager from the package-url type specifications.
         /// </summary>
         /// <seealso href="https://www.github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst"/>
-        public const string Type = null!;
+        public const string Type = "undefined";
 
         /// <summary>
         /// Initializes a new instance of the <see cref="BaseProjectManager"/> class.
--- a/src/Shared/PackageManagers/BaseProjectManager.cs
+++ b/src/Shared/PackageManagers/BaseProjectManager.cs
@@ -18,19 +18,25 @@ namespace Microsoft.CST.OpenSource.PackageManagers
     using Version = SemanticVersioning.Version;
     using PackageUrl;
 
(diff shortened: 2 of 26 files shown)


src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -18,19 +18,25 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
     using PackageUrl;
 
-    public class BaseProjectManager
+    public abstract class BaseProjectManager
     {
+        /// <summary>
+        /// The type of the project manager from the package-url type specifications.
+        /// </summary>
+        /// <seealso href="https://www.github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst"/>
+        public static readonly string Type = null!;
```
jpinz, line 27: Just wanted to note here, this **isn't** in use in this PR. It will be used in #318, and I anticipate it being useful in other places in the future as well.

src/Shared/Helpers/ArchiveHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Microsoft.CST.OpenSource.Helpers;
+
+using RecursiveExtractor;
+using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+public static class ArchiveHelper
```
brbayes-msft, line 11: Some basic unit tests for this new class would be good.

src/Shared/PackageManagers/CPANProjectManager.cs
```diff
@@ -14,6 +15,9 @@ namespace Microsoft.CST.OpenSource.PackageManagers
 
     internal class CPANProjectManager : BaseProjectManager
     {
+        /// <inheritdoc cref="BaseProjectManager.Type"/>
+        public new const string Type = "cpan";
```
brbayes-msft, line 19: Could you share some documentation as to how the const replaces the static readonly value?  I'm not familiar with this pattern.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -10,27 +10,6 @@ namespace Microsoft.CST.OpenSource.PackageManagers
 
     public static class ProjectManagerFactory
```
brbayes-msft, line 11: Are these changes to the factory necessary for this PR, or is it just general cleanup of unused static methods?

src/oss-find-squats-lib/Mutators/IMutator.cs
```diff
@@ -37,14 +38,14 @@ public IEnumerable<Mutation> Generate(PackageURL arg)
                 if (hasNamespace)
                 {
                     yield return new Mutation(
-                        mutated: arg.CreateWithNewNames(arg.Name, mutation.Mutated).ToString(),
+                        mutated: arg.CreateWithNewNames(arg.Name, HttpUtility.UrlEncode(mutation.Mutated)).ToString(),
```
brbayes-msft, line 41: Please add a test to protect against regressions here.

src/Shared/PackageManagers/CPANProjectManager.cs
jpinz: This is just what Rider told me to do lol. I can do some research into this pattern though.

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: Yeah, the first one was being used but the second was unused. Now they're both unused so I removed them both.

src/Shared/PackageManagers/CPANProjectManager.cs
gfs: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/knowing-when-to-use-override-and-new-keywords

You have to use new const to override the original const because it isn't virtual. const/static values cannot be made abstract, so this is the way to do it if you want access to the type value on a derived class.

gfs: Since the BaseProjectManager is now abstract this could be changed by just changing the Type in that class to abstract and instead of "new const" put "override".  However, then it will be an instance variable and you'll have to create a manager to determine what types it can handle.

src/oss-find-squats-lib/Mutators/IMutator.cs
jpinz: I believe this should be addressed now.

src/Shared/PackageManagers/CPANProjectManager.cs
jpinz: I did some messing around and looks like it doesn't need to be const, I will change them to `public new readonly string Type = ...` https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-modifier

src/oss-detect-cryptography/DetectCryptographyTool.cs
```diff
@@ -92,9 +92,8 @@ private static async Task Main(string[] args)
                             {
                                 targetDirectoryName = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
                             }
-                            BaseProjectManager projectManager = ProjectManagerFactory.CreateBaseProjectManager(detectCryptographyTool.HttpClientFactory, targetDirectoryName);
 
-                            string? path = await projectManager.ExtractArchive("temp", File.ReadAllBytes(target));
+                            string? path = await ArchiveHelper.ExtractArchiveAsync(targetDirectoryName, "temp", File.OpenRead(target));
```
gfs, line 96: Recommend changing this "Temp" to something more descriptive. The directory is already random and in the temp directory.

src/Shared/PackageManagers/BaseProjectManager.cs
gfs, line 27:
```suggestion
        public static readonly string Type = "Undefined";
```

Recommend not forcing this to be null.  Its still possible to access static properties of abstract classes and returning null here breaks the contract that Type is a non-null string.  No strong preference on what value to use here, `string.Empty` would be fine, `Undefined`, `Invalid` are all options that seems fine. Unfortunately I didn't see any already defined value for this type of invalid scenario in the types you posted from the purl repo.

gfs (approved):
Seems fine overall. Some minor recommendations.

src/Shared/PackageManagers/CPANProjectManager.cs
gfs: If you make it readonly it means you can't access it from a static context and you will have to create a manager to determine what it's type is.  I recommend leaving it as const.

Also you'll have problems if you have, say a CpanManager but you pass it to a function that treats it like a generic manager.
image

jpinz: I actually ended up changing the BaseProjectManager.Type to be a const instead of static readonly. Makes more sense this way,

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: Changed to `undefined`

jpinz:
@gfs I removed the `public const string Type` from the `BaseProjectManager` as there was never a case where we would be accessing `BaseProjectManager.Type`, just wanted to let you know based on the comments you left.

src/Shared/PackageManagers/CPANProjectManager.cs
jpinz: Removed `Type` from `BaseProjectManager`

gfs:
The reason you want the type property on the base manager is so that if you had a method that took a generic manager you could redetermine what type it implements. 

```C#
void SomeMethod(BaseProjectManager bpm)
{
    Switch(bpm.Type)
...
}
```

Maybe the right approach is a to make an attribute that specifies the Purl type?

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/attributes/

src/Shared/Helpers/ArchiveHelper.cs
jpinz: Done.

jpinz:
> The reason you want the type property on the base manager is so that if you had a method that took a generic manager you could redetermine what type it implements.
> 
> ```cs
> void SomeMethod(BaseProjectManager bpm)
> {
>     Switch(bpm.Type)
> ...
> }
> ```
> 
> This doesn't work if Type is not defined on base project manager.
> 
> Maybe the right approach is a to make an attribute that specifies the Purl type?
> 
> https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/attributes/

Ooh I like the idea of an attribute for the Purl type. I will go mess around with that.

jpinz:
> ```cs
> void SomeMethod(BaseProjectManager bpm)
> {
>     Switch(bpm.Type)
> ...
> }
> ```

I ended up not going with an attribute and instead what I did was added a:
`public abstract string ManagerType { get; }` to the `BaseProjectManager` and each implementation overrides it and returns `Type`

brbayes-msft (approved):

brbayes-msft (approved):

gfs:
Seems fine. 

We might need to play more with how the type is defined once we start using it but that's nbd.

gfs (approved):

PR #326, opened 2022-04-06
Refactor the ProjectManagerFactory

jpinz:
Refactor the ProjectManagerFactory to no longer use reflection for the construction of ProjectManagers.
Switch OSSGadgetLib to use a ProjectManagerFactory instead of an IHttpClientFactory, and make the change in OSSGadget as well.

### DIFF
--- a/src/Shared.CLI/Helpers/OutputBuilder/SarifOutputBuilder.cs
+++ b/src/Shared.CLI/Helpers/OutputBuilder/SarifOutputBuilder.cs
@@ -27,7 +27,7 @@ namespace Microsoft.CST.OpenSource.Shared
         /// <returns>Location list with single location object</returns>
         public static List<Location> BuildPurlLocation(PackageURL purl)
         {
-            BaseProjectManager? projectManager = ProjectManagerFactory.CreateProjectManager(purl, null);
+            BaseProjectManager? projectManager = ProjectManagerFactory.GetProjectManager(purl, null);
             if (projectManager == null)
             {
                 Logger.Debug("Cannot determine the package type");
--- a/src/Shared.CLI/Helpers/OutputBuilder/SarifOutputBuilder.cs
+++ b/src/Shared.CLI/Helpers/OutputBuilder/SarifOutputBuilder.cs
@@ -27,7 +27,7 @@ namespace Microsoft.CST.OpenSource.Shared
         /// <returns>Location list with single location object</returns>
         public static List<Location> BuildPurlLocation(PackageURL purl)
         {
-            BaseProjectManager? projectManager = ProjectManagerFactory.CreateProjectManager(purl, null);
+            BaseProjectManager? projectManager = ProjectManagerFactory.ConstructPackageManager(purl, null);
             if (projectManager == null)
             {
                 Logger.Debug("Cannot determine the package type");
--- a/src/Shared.CLI/Helpers/OutputBuilder/SarifOutputBuilder.cs
+++ b/src/Shared.CLI/Helpers/OutputBuilder/SarifOutputBuilder.cs
@@ -27,7 +27,7 @@ namespace Microsoft.CST.OpenSource.Shared
         /// <returns>Location list with single location object</returns>
         public static List<Location> BuildPurlLocation(PackageURL purl)
         {
-            BaseProjectManager? projectManager = ProjectManagerFactory.GetProjectManager(purl, null);
+            BaseProjectManager? projectManager = ProjectManagerFactory.CreateProjectManager(purl, null);
             if (projectManager == null)
             {
                 Logger.Debug("Cannot determine the package type");
--- a/src/Shared.CLI/Helpers/PackageDownloader.cs
+++ b/src/Shared.CLI/Helpers/PackageDownloader.cs
@@ -22,10 +22,10 @@ namespace Microsoft.CST.OpenSource
         /// Constructor - creates a class object for downloading packages.
         /// </summary>
         /// <param name="purl">The package to download.</param>
-        /// <param name="httpClientFactory">The <see cref="IHttpClientFactory"/> to use to make the client to use when downloading.</param>
+        /// <param name="projectManagerFactory">The <see cref="ProjectManagerFactory"/> to use to get the project managers.</param>
         /// <param name="destinationDir">The directory where the package needs to be downloaded to.</param>
         /// <param name="doCaching">Check and use the cache if it exists - create if not.</param>
-        public PackageDownloader(PackageURL purl, IHttpClientFactory? httpClientFactory, string? destinationDir = null, bool doCaching = false)
+        public PackageDownloader(PackageURL purl, ProjectManagerFactory projectManagerFactory, string? destinationDir = null, bool doCaching = false)
         {
             if (purl == null)
             {
@@ -47,7 +47,7 @@ namespace Microsoft.CST.OpenSource
                 destinationDirectory = destinationDir;
             }
 
-            packageManager = ProjectManagerFactory.CreateProjectManager(purl, httpClientFactory, destinationDirectory);
+            packageManager = projectManagerFactory.CreateProjectManager(purl, destinationDirectory);
             if (packageManager == null)
             {
                 // Cannot continue without a package manager.
--- a/src/Shared.CLI/Helpers/PackageDownloader.cs
+++ b/src/Shared.CLI/Helpers/PackageDownloader.cs
@@ -25,7 +25,7 @@ namespace Microsoft.CST.OpenSource
         /// <param name="projectManagerFactory">The <see cref="ProjectManagerFactory"/> to use to get the project managers.</param>
         /// <param name="destinationDir">The directory where the package needs to be downloaded to.</param>
         /// <param name="doCaching">Check and use the cache if it exists - create if not.</param>
-        public PackageDownloader(PackageURL purl, ProjectManagerFactory projectManagerFactory, string? destinationDir = null, bool doCaching = false)
+        public PackageDownloader(PackageURL purl, ProjectManagerFactory? projectManagerFactory = null, string? destinationDir = null, bool doCaching = false)
         {
             if (purl == null)
             {
@@ -47,7 +47,7 @@ namespace Microsoft.CST.OpenSource
                 destinationDirectory = destinationDir;
             }
 
-            packageManager = projectManagerFactory.CreateProjectManager(purl, destinationDirectory);
+            packageManager = (projectManagerFactory ?? new ProjectManagerFactory()).CreateProjectManager(purl, destinationDirectory);
             if (packageManager == null)
             {
                 // Cannot continue without a package manager.
--- a/src/Shared.CLI/Helpers/PackageDownloader.cs
+++ b/src/Shared.CLI/Helpers/PackageDownloader.cs
@@ -25,7 +25,7 @@ namespace Microsoft.CST.OpenSource
         /// <param name="projectManagerFactory">The <see cref="ProjectManagerFactory"/> to use to get the project managers.</param>
         /// <param name="destinationDir">The directory where the package needs to be downloaded to.</param>
         /// <param name="doCaching">Check and use the cache if it exists - create if not.</param>
-        public PackageDownloader(PackageURL purl, ProjectManagerFactory? projectManagerFactory = null, string? destinationDir = null, bool doCaching = false)
+        public PackageDownloader(PackageURL purl, ProjectManagerFactory projectManagerFactory, string? destinationDir = null, bool doCaching = false)
         {
             if (purl == null)
             {
@@ -47,7 +47,7 @@ namespace Microsoft.CST.OpenSource
                 destinationDirectory = destinationDir;
             }
 
-            packageManager = (projectManagerFactory ?? new ProjectManagerFactory()).CreateProjectManager(purl, destinationDirectory);
+            packageManager = projectManagerFactory.CreateProjectManager(purl, destinationDirectory);
             if (packageManager == null)
             {
                 // Cannot continue without a package manager.
--- a/src/Shared.CLI/Helpers/RepoSearch.cs
+++ b/src/Shared.CLI/Helpers/RepoSearch.cs
@@ -2,14 +2,11 @@
 
 namespace Microsoft.CST.OpenSource.Shared
 {
-    using Microsoft.CST.OpenSource.Helpers;
     using Microsoft.CST.OpenSource.PackageManagers;
-    using NLog;
     using PackageUrl;
     using System;
     using System.Collections.Generic;
     using System.Linq;
-    using System.Net.Http;
     using System.Threading.Tasks;
 
     /// <summary>
@@ -28,16 +25,16 @@ namespace Microsoft.CST.OpenSource.Shared
     /// </summary>
     public class RepoSearch
     {
-        public RepoSearch(IHttpClientFactory httpClientFactory)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RepoSearch"/> class.
+        /// </summary>
+        /// <param name="projectManagerFactory">The <see cref="ProjectManagerFactory"/> to generate the project managers with.</param>
+        public RepoSearch(ProjectManagerFactory? projectManagerFactory = null)
         {
-            HttpClientFactory = httpClientFactory;
+            _projectManagerFactory = projectManagerFactory ?? new ProjectManagerFactory();
         }
 
-        public RepoSearch() : this(new DefaultHttpClientFactory())
-        {
-        }
-
-        private IHttpClientFactory HttpClientFactory { get; }
+        private ProjectManagerFactory _projectManagerFactory { get; }
 
         /// <summary>
         ///     try to resolve the source code for an npm package through different means
@@ -62,7 +59,7 @@ namespace Microsoft.CST.OpenSource.Shared
             Logger.Debug("Searching for source code for: {0}", purlNoVersion.ToString());
 
             // Use reflection to find the correct downloader class
-            BaseProjectManager? projectManager = ProjectManagerFactory.CreateProjectManager(purl, HttpClientFactory);
+            BaseProjectManager? projectManager = _projectManagerFactory.CreateProjectManager(purl);
 
             if (projectManager != null)
             {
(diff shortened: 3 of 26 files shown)


src/Shared.CLI/Helpers/OutputBuilder/SarifOutputBuilder.cs
```diff
@@ -27,7 +27,7 @@ public SarifOutputBuilder(SarifVersion version)
         /// <returns>Location list with single location object</returns>
         public static List<Location> BuildPurlLocation(PackageURL purl)
         {
-            BaseProjectManager? projectManager = ProjectManagerFactory.CreateProjectManager(purl, null);
+            BaseProjectManager? projectManager = ProjectManagerFactory.GetProjectManager(purl, null);
```
gfs, line 30: Why rename Get instead of Create? Get implies to me that the object is somehow a singleton/cached. But this method always eventually ctors (at least for the default managers, your `CreatePackageManager` method could return a singleton if you desired).

src/Shared.CLI/Helpers/PackageDownloader.cs
```diff
@@ -22,10 +22,10 @@ public class PackageDownloader
         /// Constructor - creates a class object for downloading packages.
         /// </summary>
         /// <param name="purl">The package to download.</param>
-        /// <param name="httpClientFactory">The <see cref="IHttpClientFactory"/> to use to make the client to use when downloading.</param>
+        /// <param name="projectManagerFactory">The <see cref="ProjectManagerFactory"/> to use to get the project managers.</param>
         /// <param name="destinationDir">The directory where the package needs to be downloaded to.</param>
         /// <param name="doCaching">Check and use the cache if it exists - create if not.</param>
-        public PackageDownloader(PackageURL purl, IHttpClientFactory? httpClientFactory, string? destinationDir = null, bool doCaching = false)
+        public PackageDownloader(PackageURL purl, ProjectManagerFactory projectManagerFactory, string? destinationDir = null, bool doCaching = false)
```
gfs, line 28: Its inconsistent that this requires a ProjectManagerFactory isntead of being able to intantiate a default one.

src/Shared.CLI/Helpers/RepoSearch.cs
```diff
@@ -28,16 +25,16 @@ namespace Microsoft.CST.OpenSource.Shared
...
-        }
-
-        public RepoSearch() : this(new DefaultHttpClientFactory())
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RepoSearch"/> class.
+        /// </summary>
+        /// <param name="projectManagerFactory">The <see cref="ProjectManagerFactory"/> to generate the project managers with.</param>
+        public RepoSearch(ProjectManagerFactory? projectManagerFactory = null)
         {
+            _projectManagerFactory = projectManagerFactory ?? new ProjectManagerFactory();
```
gfs, line 34: This is the pattern we've generally adopted when provided a null client/'manager.

src/Shared.CLI/OSSGadget.cs
```diff
@@ -21,11 +20,7 @@ public class OSSGadget : OssGadgetLib
         public static string ToolName { get => GetToolName() ?? ""; }
         public static string ToolVersion { get => GetToolVersion() ?? ""; }
 
-        public OSSGadget(IHttpClientFactory httpClientFactory) : base(httpClientFactory)
-        {
-        }
-
-        public OSSGadget() : base()
```
gfs, line 28: Why remove this ctor?

gfs:
```c#
public OSSGadget() : base()
```
Is still valid - OSsGadgetLib has a matching base ctor.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,153 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
+    public class ProjectManagerFactory
     {
         /// <summary>
-        /// Get an appropriate project manager for package given its PackageURL.
+        /// The delegate to use when defining the function to create a project manager with.
+        /// The only runtime parameter we need is the destination directory. Everything else can be defined in the constructor.
         /// </summary>
-        /// <param name="purl">The <see cref="PackageURL"/> for the package to create the project manager for.</param>
-        /// <param name="httpClientFactory"> The <see cref="IHttpClientFactory"/> for the project manager to use for making Http Clients to make web requests.</param>
-        /// <param name="destinationDirectory">The directory to use to store any downloaded packages.</param>
```
gfs, line 18: Restore doc comment

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,153 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
-        /// <returns> BaseProjectManager object </returns>
```
gfs, line 19: Restore doc comment, add "or null if unable to Construct"

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,153 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
+
+        /// <summary>
+        /// The dictionary of project managers.
+        /// </summary>
+        private readonly Dictionary<string, ConstructProjectManager> _projectManagers;
+
+        /// <summary>
+        /// Initializes a new instance of a <see cref="ProjectManagerFactory"/>.
+        /// </summary>
+        /// <param name="overrideManagers">Use a custom set of managers to override the defaults with.</param>
```
gfs, line 27:
```suggestion
        /// <param name="overrideManagers">If provided, will override each matching manager in the defaults - or add if not present in default.</param>
```
Needs to be clear it doesn't override all defaults it only overrides the specified ones. Not sure if this is the best way to write it.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,153 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
-               .Where(type => type.IsSubclassOf(typeof(BaseProjectManager))));
-            }
-            // Use reflection to find the correct package management class
-            Type? downloaderClass = projectManagers
-               .Where(type => type.Name.Equals($"{purl.Type}ProjectManager",
-                                               StringComparison.InvariantCultureIgnoreCase))
-               .FirstOrDefault();
-
-            if (downloaderClass != null)
+            overrideManagers.ToList().ForEach(pair => SetManager(pair.Key, pair.Value));
```
gfs, line 31: This seems like an unnecessary allocation.

```suggestion
            foreach(var pair in overrideManagers)
            {
                SetManager(pair.Key, pair.Value);
            }
```

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,153 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
+            _projectManagers[lookup] = generator;
+        }
+
+        public void ClearManagers()
+        {
+            _projectManagers.Clear();
+        }
+
+        /// <summary>
+        ///
```
gfs, line 59:
```suggestion
        /// Construct the default set of managers (optionally using a specified IHttpClientFactory) that ship with OSSGadget.
```

gfs:
Nit: PR Description. This is no longer using reflection for construction.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,153 @@ namespace Microsoft.CST.OpenSource.PackageManagers
     using System.Linq;
     using System.Net.Http;
 
-    public static class ProjectManagerFactory
+    public class ProjectManagerFactory
     {
         /// <summary>
-        /// Get an appropriate project manager for package given its PackageURL.
+        /// The delegate to use when defining the function to create a project manager with.
```
gfs, line 14:
```suggestion
        /// This delegate takes the path to operate a project manager in and returns a ProjectManager.
```

Add another line describing how you would actually use this by constructing a dictionary?

gfs (changes requested):
Minor changes. The important one is to note the allocation implications of .ToList.

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: I don't think this applies here anymore? I did still update the documentation.

src/Shared.CLI/Helpers/RepoSearch.cs
jpinz: This was in response to https://github.com/microsoft/OSSGadget/pull/326#discussion_r844243432 right? I don't think I have anything to change here but I could be reading what you commented wrong.

src/Shared.CLI/Helpers/RepoSearch.cs
```diff
@@ -28,16 +25,16 @@ namespace Microsoft.CST.OpenSource.Shared
...
         }
```
pmalmsten, line 35: Convention question regarding these helper / tool classes: how often do we call these constructors without providing a project manager value (i.e. this class is at the top of a dependency graph), vs. how often do these constructors get calls from other helper/tool classes which need to pass the same factory object reference (i.e. this class is not at the top of a dependency graph)?

I ask because I wonder whether we should just make the `ProjectManagerFactory` dependency non-nullable and required for all of the helper/tool classes. If we frequently re-use these classes from within other such classes, making this parameter required would allow the compiler to remind us that we need to pass a value in instead of accidentally not passing a value and using the default `ProjectManagerFactory` instance where we shouldn't.

pmalmsten (commented):
Generally looks straightforward, only a few small notes.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,153 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
+
+        private void SetManager(string lookup, ConstructProjectManager generator)
+        {
+            _projectManagers[lookup] = generator;
+        }
+
+        public void ClearManagers()
+        {
+            _projectManagers.Clear();
+        }
```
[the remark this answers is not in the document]
jpinz: SetManager will be used when we are overriding a constructor, such as for unit testing NuGet we want to override the constructor with a mocked ManagerPackageActions (Not yet implemented)

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: I added an example and a pointer to the CreateDefaultManagers method

src/Shared/PackageManagers/ProjectManagerFactory.cs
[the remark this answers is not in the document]
pmalmsten: > SetManager will be used when we are overriding a constructor, such as for unit testing NuGet we want to override the constructor with a mocked ManagerPackageActions

Could that be accomplished by passing the overrides into one of the constructors of this class instead? It looks like there is already a constructor on this class that takes overrides?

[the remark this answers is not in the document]
gfs: We can remove these methods if we change the constructor to take a full set instead of a subset to replace, and then in fact declare that it is backed by an immutable dictionary.

Since we have a "GetDefault" that is public the caller can get the default and then replace what they want to replace and then call the ctor.

src/Shared.CLI/Helpers/OutputBuilder/SarifOutputBuilder.cs
jpinz: I switched it to `ConstructPackageManager` and added the `overrideManagers` dict and the `destinationDirectory` as optional parameters. That make sense?

src/Shared.CLI/Helpers/RepoSearch.cs
jpinz: The helpers it should never be null, but I think it's ok for the tool classes to not take in a `ProjectManagerFactory` and it be null by default. We only would override those in unit test cases, no one consumes the CLI tools and would have to worry about it being the default factory.

gfs: > I ask because I wonder whether we should just make the ProjectManagerFactory dependency non-nullable and required for all of the helper/tool classes.

For Helper classes I think this is fine.  The only references I see to creating this without project manager factory specified is in tests, where it probably especially makes sense to use a mock instead of defaults.

For the OSSGadget's themselves I do not agree.  I think more users than not will not make any configuration changes to the default manager set and we should keep the parameterless ctors.

src/Shared/PackageManagers/ProjectManagerFactory.cs
gfs: It is a different method but the parameter is named the same. The new method doesn't have a comment for the parameter.

https://github.com/microsoft/OSSGadget/blob/71cae70070f34a783dcefea12fdd7a351cfcd60f/src/Shared/PackageManagers/ProjectManagerFactory.cs#L13-L17

src/Shared.CLI/Helpers/RepoSearch.cs
jpinz: I made it non-nullable in RepoSearch and PackageDownloader.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,153 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="httpClientFactoryParam">The <see cref="IHttpClientFactory"/> to use in the project managers.</param>
+        /// <returns>A dictionary of the default managers with associated constructors.</returns>
+        private static Dictionary<string, ConstructProjectManager> CreateDefaultManagers(IHttpClientFactory? httpClientFactoryParam = null)
```
gfs, line 89: This needs to be public if we remove the specific get/set methods so you can modify the default set and then pass it in.

src/Shared/PackageManagers/ProjectManagerFactory.cs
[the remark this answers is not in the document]
jpinz: I removed UnsetManager and ClearManagers. I can't think of a scenario we would use them.

src/Shared.CLI/Helpers/RepoSearch.cs
```diff
@@ -62,7 +59,7 @@ public RepoSearch() : this(new DefaultHttpClientFactory())
             Logger.Debug("Searching for source code for: {0}", purlNoVersion.ToString());
 
             // Use reflection to find the correct downloader class
```
gfs, line 61:
```suggestion
```

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: I removed the UnsetManager and ClearManagers methods. I can't think of a case where we would need either of those, and I also can't think of a case where we would need a getter outside of the `ProjectManagerFactory` class.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -8,55 +8,171 @@ namespace Microsoft.CST.OpenSource.PackageManagers
...
+        /// </param>
+        /// <param name="httpClientFactory">The <see cref="IHttpClientFactory"/> to use in the project managers.</param>
+        public ProjectManagerFactory(Dictionary<string, ConstructProjectManager> overrideManagers, IHttpClientFactory? httpClientFactory = null) : this(httpClientFactory)
         {
-            if (projectManagers.Count == 0)
+            foreach((string? type, ConstructProjectManager? ctor) in overrideManagers)
             {
-                projectManagers.AddRange(typeof(BaseProjectManager).Assembly.GetTypes()
-               .Where(type => type.IsSubclassOf(typeof(BaseProjectManager))));
+                SetManager(type, ctor);
```
gfs, line 43: Make sure those values aren't null before adding them?

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: Ahh, I did add that. Should be good.

gfs:
That makes sense to me.

jpinz:
> That makes sense to me.
> […]

@gfs just made the change and added examples in a new ProjectManagerFactoryTests class.

src/oss-tests/ProjectManagerTests/ProjectManagerFactoryTests.cs
```diff
@@ -0,0 +1,113 @@
...
+    }
+    
+    [DataTestMethod]
+    [DataRow(null)] // Add nothing, remove nothing.
+    [DataRow(0)] // Add a "test" type with the NuGetProjectManager, remove nothing.
+    [DataRow(1)] // Add a "nuget" type with the NuGetProjectManager, overriding the default, remove nothing.
+    [DataRow(2)] // Add a "nuget" type but use an NPMProjectManager, overriding the default, remove nothing.
+    [DataRow(3)] // Add "nuget", "npm" and "pypi" types, overriding the defaults, remove nothing.
+    [DataRow(null, "nuget")] // Add nothing, remove "nuget".
+    [DataRow(null, "nuget", "npm")] // Add nothing, remove "nuget" & "npm".
```
pmalmsten, line 71: These data rows are rather convoluted. I wonder if in this case, it would be easier/clearer to just write these as separate tests and not try to parameterize them?

(Personally, when I see comments like these, it makes me think how we might use one or more methods with appropriate names to reduce or eliminate the need for comments to explain what the code is doing.)

jpinz: Yeah definitely kind of a mess, just wanted to get some tests going. I will clean it up now.

jpinz: @pmalmsten cleaned it up a lot I think.

gfs:
In meetings for the rest of the day. May take me until tomorrow to  get to reviews.

pmalmsten (approved):
Looks great, thanks for all of the improvements!

src/Shared.CLI/Helpers/RepoSearch.cs
gfs: This was just extra context for another comment. No actoin needed on this line.

gfs (approved):
Great job! This looks very clean.

PR #327, opened 2022-04-09
Implement IManagerPackageActions and related work.

jpinz:
Some of the logic pertaining to project managers can be extracted into libraries.  So the `IManagerPackageActions` interface was created which defines methods to download a package version, get all versions of a package, get the latest version of a package, and get a package's metadata. Currently, the only implementation is for NuGet which now uses the `NuGet.Protocol` library, I created a `NuGetPackageActions` class to implement the interface and use `NuGet.Protocol` to implement the methods defined in `IManagerPackageActions`


jpinz:
The download tests are failing for NuGet. I'm OOF until wednesday, just wanted to get this out for review so hopefully I have some comments to address on wednesday.

src/Shared/PackageActions/NuGetPackageActions.cs
```diff
@@ -0,0 +1,169 @@
...
+        bool doExtract,
+        bool cached = false,
+        CancellationToken cancellationToken = default)
+    {
+        FindPackageByIdResource resource = await _sourceRepository.GetResourceAsync<FindPackageByIdResource>();
+
+        // Construct the path for the nupkg file.
+        string filePath = Path.ChangeExtension(Path.Join(topLevelDirectory, targetPath), Path.GetExtension(targetPath) + ".nupkg");
+
+        // Create a new memory stream to populate with the .nupkg.
```
pmalmsten, line 42: Nit: Is this comment still accurate? Looks like it might be out of date.

src/oss-tests/FindSquatsTests.cs
```diff
@@ -28,7 +28,7 @@ namespace Microsoft.CST.OpenSource.Tests
     public class FindSquatsTest
     {
         [DataTestMethod]
-        [DataRow("pkg:nuget/Microsoft.CST.OAT", false)]
+        // [DataRow("pkg:nuget/Microsoft.CST.OAT", false)] TODO: Figure out why this is failing on the ADO pipeline.
```
gfs, line 31: Is it possible you're getting rate limited? Running these tests takes a bit of time now due to the number of requests. it looks like it is generating 300+ candidates for this case - which is the only nuget case.  The npm cases generate 90~ and 400~ respectively.

I think this test should be refactored to perhaps validate the mutations that were generated rather than actually querying the repository - or mock those requests.

gfs (changes requested):
Looks good overall. The majority of this code has been through previous reviews in a separate PR.

Main concern is with the findsquats test which produces almost 1000 network requests for the 3 cases and takes quite a while to execute. I think we don't need the test to actually check the repository - also has the potential to fail if a squat does show up, but the packages chosen were arbitrary.

Also should the Base manager's method be renamed `...Async` per c# convention?

src/Shared/PackageActions/NuGetPackageActions.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        IEnumerable<NuGetVersion> versions = await resource.GetAllVersionsAsync(
+            packageUrl.Name,
+            _sourceCacheContext,
+            _logger, 
+            cancellationToken);
+
+        return versions
+            .Where(v => includePrerelease || !v.IsPrerelease)
+            .Select(v => v.ToString()).Reverse();
```
pmalmsten, line 124: Tiny nit: when working with fluent APIs like IEnumerable methods, I find it a bit easier to read when all of the fluent method calls have the same indent level. In this case, the `.Reverse()` to the right of the `Select` did not jump out to me until I thought more carefully about the sort order contract for this API: 
```suggestion
        return versions
            .Where(v => includePrerelease || !v.IsPrerelease)
            .Select(v => v.ToString())
            .Reverse();
```

src/Shared/PackageActions/NuGetPackageActions.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// <inheritdoc />
+    public async Task<IEnumerable<string>> GetAllVersionsAsync(
+        PackageURL packageUrl,
+        bool includePrerelease = true,
+        bool useCache = true,
+        CancellationToken cancellationToken = default)
+    {
+        FindPackageByIdResource resource = await _sourceRepository.GetResourceAsync<FindPackageByIdResource>();
+
+        IEnumerable<NuGetVersion> versions = await resource.GetAllVersionsAsync(
```
pmalmsten, line 116: What do you think about including the sort order as part of the name for this variable? Perhaps something like `versionsOldestFirst`?

src/Shared/PackageActions/NuGetPackageActions.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// <inheritdoc />
+    public async Task<string?> GetLatestVersionAsync(
+        PackageURL packageUrl,
+        bool includePrerelease = false,
+        bool useCache = true,
+        CancellationToken cancellationToken = default)
+    {
+        FindPackageByIdResource resource = await _sourceRepository.GetResourceAsync<FindPackageByIdResource>();
+
+        IEnumerable<NuGetVersion> versions = await resource.GetAllVersionsAsync(
```
pmalmsten, line 136: Same thought as above - what do you think about including the sort order as part of the name for this variable? Perhaps something like `versionsOldestFirst`?

src/oss-tests/ProjectManagerTests/ProjectManagerFactoryTests.cs
```diff
@@ -64,7 +64,7 @@ public void DefaultSucceeds()
     [TestMethod]
     public void AddTestManagerSucceeds()
     {
-        _managerOverrides["test"] = directory => new NuGetProjectManager(_mockFactory.Object, directory); // Create a test type with the NuGetProjectManager.
+        _managerOverrides["test"] = directory => new NuGetProjectManager(directory, null, _mockFactory.Object); // Create a test type with the NuGetProjectManager.
```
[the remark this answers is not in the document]
jpinz: No network calls are actually being made, so mocks or not it doesn't really matter. I switched `_mockFactory` to `_httpClientFactory = new DefaultHttpClientFactory()` and removed `Moq` from this test class.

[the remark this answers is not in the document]
gfs: Ok, then no worries.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -285,7 +285,7 @@ public virtual Task<IEnumerable<string>> DownloadVersion(PackageURL purl, bool d
         /// <param name="useCache">If the cache should be used when looking for the versions.</param>
         /// <param name="includePrerelease">If pre-release versions should be included.</param>
         /// <returns> A list of package version numbers.</returns>
-        public virtual Task<IEnumerable<string>> EnumerateVersions(PackageURL purl, bool useCache = true, bool includePrerelease = true)
+        public virtual Task<IEnumerable<string>> EnumerateVersionsAsync(PackageURL purl, bool useCache = true, bool includePrerelease = true)
```
gfs, line 288: needs async marker after virtual

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -272,7 +272,7 @@ public static IEnumerable<string> SortVersions(IEnumerable<string> versionList)
         /// </summary>
         /// <param name="purl">PackageURL to download</param>
         /// <returns>Paths (either files or directory names) pertaining to the downloaded files.</returns>
-        public virtual Task<IEnumerable<string>> DownloadVersion(PackageURL purl, bool doExtract, bool cached = false)
+        public virtual Task<IEnumerable<string>> DownloadVersionAsync(PackageURL purl, bool doExtract, bool cached = false)
```
gfs, line 275: needs async marker after virtual

src/oss-tests/FindSquatsTests.cs
```diff
@@ -24,7 +28,7 @@ namespace Microsoft.CST.OpenSource.Tests
...
```
pmalmsten, line 31: Should this get figured out before merging this PR?

jpinz:
@gfs If we mark the `BaseProjectManager` async methods as `async` they'll warn us unless we suppress the warnings, or have an await in this implementation. 
image

src/oss-tests/Helpers/PackageActionsHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Threading;
+
+public static class PackageActionsHelper<T> where T : IManagerPackageVersionMetadata
+{
+    /// <summary>
+    /// Set up a mock of <see cref="IManagerPackageActions{T}"/> for this test run.
+    /// </summary>
+    /// <param name="purl">The <see cref="PackageURL"/> to use when configuring the mocked calls for this manager.</param>
+    /// <param name="metadata">The <see cref="T"/> to use when returning the call to <see cref="IManagerPackageActions{T}.GetMetadataAsync"/>.</param>
+    /// <param name="versions">The list of versions to return when mocking the call
```
pmalmsten, line 21: Should we document the expected sort order for this enumerable?

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -307,15 +307,15 @@ public virtual Task<IEnumerable<string>> EnumerateVersions(PackageURL purl, bool
         /// <param name="purl">The PackageURL to check.</param>
         /// <param name="useCache">If the cache should be checked for the existence of this package.</param>
         /// <returns>True if the package is confirmed to exist in the repository. False otherwise.</returns>
-        public virtual async Task<bool> PackageExists(PackageURL purl, bool useCache = true)
+        public virtual async Task<bool> PackageExistsAsync(PackageURL purl, bool useCache = true)
```
gfs, line 310: Async marker is optional with virtual - it is just a compiler hint per this dicsussion: https://stackoverflow.com/questions/25015853/is-it-ok-to-have-virtual-async-method-on-base-class. 

But should be consistent in the file.  The other virtual methods are not marked async.

jpinz: I'll probably just remove this one then

src/oss-tests/Helpers/PackageActionsHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+                mockPackageActions.Setup(actions => actions.GetAllVersionsAsync(
+                    It.Is<PackageURL>(p => p.Name.Equals(purl.Name)), 
+                    includePrerelease, 
+                    It.IsAny<bool>(), 
+                    It.IsAny<CancellationToken>()).Result)
+                    .Returns(() =>
+                    {
+                        return typeof(T) == typeof(NuGetPackageVersionMetadata)
+                            ? versionsArray.Where(v => includePrerelease || !NuGetVersion.Parse(v).IsPrerelease)
+                            : versionsArray;
```
pmalmsten, line 58: Nit: this seems a bit odd. Would you mind elaborating on why this special case filtering is here while setting up mocks? Is this because we load in sample metadata from a file or something?

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: Wait, this one has to be marked `async` as it has `await EnumerateVersionsAsync(purl, useCache)` within it. That's why it has it now and the others dont.

src/Shared.CLI/Helpers/PackageDownloader.cs
```diff
@@ -248,7 +248,7 @@ public async Task<List<PackageURL>> GetPackageVersionsToProcess(PackageURL purl)
                 {
                     try
                     {
-                        foreach (string? version in await packageManager.EnumerateVersions(purl, doCache))
+                        foreach (string? version in await packageManager.EnumerateVersionsAsync(purl, doCache))
```
gfs, line 251:
```suggestion
                        await foreach (string? version in packageManager.EnumerateVersionsAsync(purl, doCache))
```

src/oss-tests/ProjectManagerTests/NuGetProjectManagerTests.cs
```diff
@@ -23,37 +27,92 @@ public class NuGetProjectManagerTests
...
+            { "pkg:nuget/razorengine@4.2.3-beta1", Resources.razorengine_4_2_3_beta1_metadata_json },
+            { "pkg:nuget/razorengine", Resources.razorengine_latest_metadata_json },
+        };
+        
+        // Map PackageURLs to the list of versions as json.
+        private readonly IDictionary<string, string> _versions = new Dictionary<string, string>()
+        {
+            { "pkg:nuget/razorengine@4.2.3-beta1", Resources.razorengine_versions_json },
+            { "pkg:nuget/razorengine", Resources.razorengine_versions_json },
+        };
```
pmalmsten, line 43: Nit: could these be immutable?
```suggestion
        // Map PackageURLs to metadata as json.
        private readonly IDictionary<string, string> _metadata = new Dictionary<string, string>()
        {
            { "pkg:nuget/razorengine@4.2.3-beta1", Resources.razorengine_4_2_3_beta1_metadata_json },
            { "pkg:nuget/razorengine", Resources.razorengine_latest_metadata_json },
        }.ToImmutableDictionary();
        
        // Map PackageURLs to the list of versions as json.
        private readonly IDictionary<string, string> _versions = new Dictionary<string, string>()
        {
            { "pkg:nuget/razorengine@4.2.3-beta1", Resources.razorengine_versions_json },
            { "pkg:nuget/razorengine", Resources.razorengine_versions_json },
        }.ToImmutableDictionary();
```

src/Shared.CLI/Helpers/PackageDownloader.cs
gfs: Does this actually produce nulls in the enumeration?

src/Shared/PackageManagers/BaseProjectManager.cs
gfs: Ah that makes sense then.

src/oss-tests/ProjectManagerTests/NuGetProjectManagerTests.cs
```diff
@@ -23,37 +27,92 @@ public class NuGetProjectManagerTests
...
+            string purlString, 
+            int count, 
+            string latestVersion, 
+            bool includePrerelease = true)
+        {
+            PackageURL purl = new(purlString);
+            IManagerPackageActions<NuGetPackageVersionMetadata>? nugetPackageActions = PackageActionsHelper<NuGetPackageVersionMetadata>.SetupPackageActions(
+                purl,
+                JsonConvert.DeserializeObject<NuGetPackageVersionMetadata>(_metadata[purl.ToString()]),
+                JsonConvert.DeserializeObject<IEnumerable<string>>(_versions[purl.ToString()])?.Reverse(),
```
pmalmsten, line 106: Nit: would you mind elaborating on why the versions list is reversed here? Is the test data actually sorted in latest-first order, and we have to reverse it to properly emulate oldest-first order returned by the actual API/SDK?

src/Shared/PackageManagers/GitHubProjectManager.cs
```diff
@@ -140,7 +140,7 @@ public override async Task<IEnumerable<string>> DownloadVersion(PackageURL purl,
                     });
                 }
                 PackageURL purlNoVersion = new(purl!.Type, purl.Namespace, purl.Name, null, purl.Qualifiers, purl.Subpath);
-                foreach (string v in EnumerateVersions(purlNoVersion).Result)
+                foreach (string v in EnumerateVersionsAsync(purlNoVersion).Result)
```
gfs, line 143:
```suggestion
                await foreach (string v in EnumerateVersionsAsync(purlNoVersion))
```

pmalmsten (changes requested):

gfs:
> @gfs If we mark the `BaseProjectManager` async methods as `async` they'll warn us unless we suppress the warnings, or have an await in this implementation. image

@jpinz I expected as much. I think its fine to leave as is.

src/Shared.CLI/Helpers/PackageDownloader.cs
jpinz: I don't think I can do this because `EnumerateVersionsAsync` returns a `Task<IEnumberable<string>>` not a `IAsyncEnumerable<string>`

jpinz: image

jpinz:
@gfs can't respond inline to this comment for some reason: https://github.com/microsoft/OSSGadget/pull/327#discussion_r849887287
But it doesn't appear to return null strings, it can however return an empty list of strings.

src/Shared/PackageManagers/GitHubProjectManager.cs
jpinz: Same situation as https://github.com/microsoft/OSSGadget/pull/327#discussion_r849927142

jpinz: I can do `foreach (string v in await EnumerateVersionsAsync(purlNoVersion))` instead

src/Shared/PackageActions/NuGetPackageActions.cs
jpinz: `versionsAscending`?

src/oss-tests/FindSquatsTests.cs
jpinz: I think it should be fixed now that none of the tests use the internet, and it is all mocked internally as it should be.

src/oss-tests/Helpers/PackageActionsHelper.cs
jpinz: Yeah, we load it in from the direct API response as json, I just reimplement the includePrerelease logic check here as well.

src/oss-tests/ProjectManagerTests/NuGetProjectManagerTests.cs
jpinz: The test data is sorted in ascending order, and as discussed in https://github.com/microsoft/OSSGadget/pull/327#discussion_r849883922, I added to the documentation that the `PackageActionsHelper` expects the list of versions in descending order, so I reversed it here. I suppose I could have it reverse in the helper itself and keep it out of it here.
Thoughts?

src/Shared/PackageActions/NuGetPackageActions.cs
jpinz: `versionsAscending`?

src/Shared.CLI/Helpers/PackageDownloader.cs
gfs: Ah I see. Other way is fine then.

gfs:
> @gfs can't respond inline to this comment for some reason: https://github.com/microsoft/OSSGadget/pull/327#discussion_r849887287
> 
> But it doesn't appear to return null strings, it can however return an empty list of strings.

In that case ignore the suggestion for changing the await position but you should remove the ? On the  string declaration inside the forecast, the enumeration shouldn't be returning nulls. If it's empty it should skip execution of the for each body entirely. 

Unless, is this another edge case with tasks instead of AsyncEbumerable?

src/oss-tests/ProjectManagerTests/NuGetProjectManagerTests.cs
pmalmsten: Got it, that makes sense. I don't think you need to change anything - just trying to keep track of all of the order reversals involved here 😅

pmalmsten (approved):
Thanks for the updates - looks great!

gfs (approved):
👍

PR #329, opened 2022-04-15
Implement GetArtifactDownloadUris in the BaseProjectManager

jpinz:
Used to get any relevant URIs associated with a package version.
Implemented for NPM, NuGet and PyPI.
Create tests for it, and create a new PyPIProjectManagerTests class.

Also fixed scoped NPM name handling. Scoped npm purls should have the namespace prefixed with %40.


src/Shared/Extensions/PackageUrlExtension.cs
```diff
@@ -40,6 +41,21 @@ public static bool HasNamespace(this PackageURL packageUrl)
...
+    /// We want npm package's namespace to be prefixed with "%40" the percent code of "@".
+    /// </summary>
+    /// <param name="packageUrl">The <see cref="PackageURL"/> to get a formatted namespace for.</param>
+    /// <returns>The formatted namespace of <paramref name="packageUrl"/>.</returns>
+    public static string GetNamespaceFormatted(this PackageURL packageUrl)
+    {
+        if (packageUrl.Type != "npm" || packageUrl.Namespace.StartsWith("%40") || packageUrl.Namespace.StartsWith("@"))
+        {
+            return packageUrl.Namespace;
+        }
```
pmalmsten, line 54: I see that this accepts a namespace string starting with `@` - would you mind elaborating on that? Should this method only return a namespace value in one format (encoded or not encoded)?

pmalmsten (commented):
A few initial thoughts, still looking.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -267,14 +267,24 @@ public static IEnumerable<string> SortVersions(IEnumerable<string> versionList)
...
 
+        /// <summary>
+        /// Gets the relevant URI(s) to download the files related to a <see cref="PackageURL"/>.
+        /// </summary>
+        /// <param name="purl">The <see cref="PackageURL"/> to get the URI(s) for.</param>
+        /// <returns>Links to the relevant URI(s).</returns>
+        public virtual IEnumerable<string> GetArtifactDownloadUris(PackageURL purl)
+        {
+            throw new NotImplementedException("BaseProjectManager does not implement GetArtifactDownloadUri.");
+        }
```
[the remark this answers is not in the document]
jpinz: Just created an issue #330 let's start a discussion there?

src/Shared/Extensions/PackageUrlExtension.cs
```diff
@@ -40,6 +41,21 @@ public static bool HasNamespace(this PackageURL packageUrl)
...
     }
     
+    /// <summary>
+    /// We want npm package's namespace to be prefixed with "%40" the percent code of "@".
+    /// </summary>
+    /// <param name="packageUrl">The <see cref="PackageURL"/> to get a formatted namespace for.</param>
+    /// <returns>The formatted namespace of <paramref name="packageUrl"/>.</returns>
+    public static string GetNamespaceFormatted(this PackageURL packageUrl)
+    {
+        if (packageUrl.Type != "npm" || packageUrl.Namespace.StartsWith("%40") || packageUrl.Namespace.StartsWith("@"))
```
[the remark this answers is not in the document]
jpinz: Good catch. I'll fix it.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -190,7 +202,7 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
 
         public override Uri GetPackageAbsoluteUri(PackageURL purl)
         {
-            return new Uri($"{ENV_NPM_API_ENDPOINT}/{purl?.Name}");
+            return new Uri(ENV_NPM_API_ENDPOINT.EnsureTrailingSlash() + (purl.HasNamespace() ? $"{purl.GetNamespaceFormatted()}/{purl.Name}" : purl.Name));
```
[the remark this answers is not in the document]
jpinz: I just had realized this wouldn't work for npm packages with a scope/namespace so I tried to fix it. What is the URI we want this method to return?
https://www.npmjs.com/package/lodash or https://registry.npmjs.org/lodash
https://www.npmjs.com/package/%40angular/core or https://registry.npmjs.org/%40angular/core

Or do we want this to return the frontend (html/css) URI or the API (json) URI?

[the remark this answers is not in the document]
gfs: I think this goes to the other discussion we want. For automation style things the json is definitely what we want. But we might also want to return/get the other url in some manner.

I think the json is the right answer for this api as it is.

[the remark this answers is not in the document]
jpinz: So then right now that's what this method returns. it's also what it was doing before the change in this PR, but it wouldn't have worked due to the namespace being ignored.

src/Shared/PackageManagers/BaseProjectManager.cs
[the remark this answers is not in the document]
jpinz: Tried a potential solution in the latest update https://github.com/microsoft/OSSGadget/pull/329/commits/dc7968111f2e42819de7a8b105e402bfbd83df62

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -267,14 +267,24 @@ public static IEnumerable<string> SortVersions(IEnumerable<string> versionList)
             return versionPartsList.Select(s => string.Join("", s));
         }
 
+        /// <summary>
+        /// Gets the relevant URI(s) to download the files related to a <see cref="PackageURL"/>.
+        /// </summary>
+        /// <param name="purl">The <see cref="PackageURL"/> to get the URI(s) for.</param>
+        /// <returns>Links to the relevant URI(s).</returns>
+        public virtual IEnumerable<string> GetArtifactDownloadUris(PackageURL purl)
```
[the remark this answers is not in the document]
jpinz: Check https://github.com/microsoft/OSSGadget/pull/329/commits/dc7968111f2e42819de7a8b105e402bfbd83df62 to see what you think. I created an `ArtifactUri` struct that contains the Type, Uri, and Extension.

src/Shared/Extensions/PackageUrlExtension.cs
jpinz: I believe this has been resolved. What the method does now is if the namespace isn't from an NPM package, or if it has `%40` at the beginning of the string, then it just returns the namespace.
If it starts with `@` it removes the `@` at the beginning of the namespace and replaces it with `%40`.
Otherwise, if it is a scoped NPM package and the namespace doesn't start with `@` or `%40` it prepends the namespace with `%40`.

src/Shared/Model/ArtifactUri.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+
+namespace Microsoft.CST.OpenSource.Model;
+
+using System;
+using System.IO;
+
+public struct ArtifactUri
```
pmalmsten, line 8: Nit: Do you think this could be a [`readonly record struct`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record)?

src/Shared/Model/ArtifactUri.cs
```diff
@@ -0,0 +1,47 @@
...
+        Uri = uri;
+        Extension = extension ?? Path.GetExtension(uri.AbsolutePath);
+    }
+
+    public ArtifactUri(ArtifactType type, string uri, string? extension = null)
+    {
+        Type = type;
+        Uri = new Uri(uri);
+        Extension = extension ?? Path.GetExtension(uri);
+    }
```
pmalmsten, line 22: Should one of these constructors call the other one, to minimize duplication? For example, maybe the constructor that takes a string could turn the string into a `Uri` as part of calling the `Uri` constructor overload?

src/Shared/Model/ArtifactUri.cs
```diff
@@ -0,0 +1,47 @@
...
+        Nuspec,
+
+        // PyPI specific
+        Wheel,
+
+        // Maven specific
+        Pom,
+        Jar,
+        Sources,
+    }
```
pmalmsten, line 46: I don't want to get too clever if it complicates things, but I wonder whether the `ArtifactUri` struct could be generic across the enum type? So instead of defining `ArtifactType` here, the `Type` field on this class would have a generic `enum` type?

That would let the NuGet-specific classes return NuGet specific artifact type values, Maven-specific classes could return Maven-specific values, etc.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -131,7 +133,7 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
                 string packageName = purl.GetFullName();
                 HttpClient httpClient = CreateHttpClient();
 
-                JsonDocument doc = await GetJsonCache(httpClient, $"{ENV_NPM_API_ENDPOINT}/{packageName}", useCache);
+                JsonDocument doc = await GetJsonCache(httpClient, $"{ENV_NPM_API_ENDPOINT}/{purl.GetFullName(true)}", useCache);
```
pmalmsten, line 136: Nit: I personally find boolean values without names to be less convenient to understand - how about using named arguments style?

```suggestion
                JsonDocument doc = await GetJsonCache(httpClient, $"{ENV_NPM_API_ENDPOINT}/{purl.GetFullName(encoded: true)}", useCache);
```

src/Shared/Model/ArtifactUri.cs
jpinz: I hadn't thought about that, I like it.

pmalmsten: This is closer, but as written right now, the compiler doesn't offer users of this field any assistance in terms of what the value might be.

This is more along the lines of what I was getting at:

```c#
public readonly record struct ArtifactUri<T> where T : System.Enum
{
    // <snip />
    T Type {get;}
   
   // <snip />
}
```

That way, the compiler can offer assistance to users of this class to know what possible Enum values could be returned. The tradeoff being that we would have to propagate the generic parameter through appropriate places.

What do you think?

src/Shared/Extensions/PackageUrlExtension.cs
```diff
@@ -40,6 +41,26 @@ public static bool HasNamespace(this PackageURL packageUrl)
         return packageUrl.Namespace.IsNotBlank();
     }
     
+    /// <summary>
+    /// We want npm package's namespace to be prefixed with "%40" the percent code of "@".
+    /// </summary>
+    /// <param name="packageUrl">The <see cref="PackageURL"/> to get a formatted namespace for.</param>
+    /// <returns>The formatted namespace of <paramref name="packageUrl"/>.</returns>
+    public static string GetNamespaceFormatted(this PackageURL packageUrl)
```
bhuwan-agarwal, line 49: For my learning, does this logic apply only to npm or are their other package managers using `@` in their namespace. It'll be helpful to put an example of a formatted namespace in the comments to better understand the logic here.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -190,7 +203,7 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
 
         public override Uri GetPackageAbsoluteUri(PackageURL purl)
```
bhuwan-agarwal, line 204: What's the difference between this Url and artifact download url?

bhuwan-agarwal: Found this in the [`BaseProjectManager`](https://github.com/microsoft/OSSGadget/blob/main/src/Shared/PackageManagers/BaseProjectManager.cs#L350). Maybe the method name can be changed later on as part of other enhancements.

src/Shared/Extensions/PackageUrlExtension.cs
jpinz: As far as I'm aware, yes it is just for NPM. I can add a remark saying that if the `packageUrl` isn't `npm` then it just returns the namespace as is.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -37,6 +38,18 @@ public NPMProjectManager(string destinationDirectory) : base(destinationDirector
         {
         }
 
+        /// <inheritdoc />
+        public override IEnumerable<ArtifactUri> GetArtifactDownloadUris(PackageURL purl)
```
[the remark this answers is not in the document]
jpinz: I can implement that pretty easily yeah. I think it makes sense.

src/Shared/PackageManagers/NPMProjectManager.cs
jpinz: This brings you to, for example, https://registry.npmjs.org/lodash, vs https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz for the artifactUri. It doesn't link to a download uri, and it doesn't specify a version.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -44,6 +45,19 @@ public class NuGetProjectManager : TypedManager<NuGetPackageVersionMetadata>
...
             GetRegistrationEndpointAsync().Wait();
         }
+        
+        /// <inheritdoc />
+        public override async IAsyncEnumerable<NuGetBaseArtifactUri> GetArtifactDownloadUrisAsync(PackageURL purl)
+        {
+            string feedUrl = (purl.Qualifiers?["repository_url"] ?? NUGET_DEFAULT_CONTENT_ENDPOINT).EnsureTrailingSlash();
+
+            string artifactUri = $"{feedUrl}{purl.Name.ToLower()}/{purl.Version}/{purl.Name.ToLower()}.{purl.Version}.nupkg";
+            HttpClient httpClient = CreateHttpClient();
```
gfs, line 55: Please remove this. Adding a network call and potentially downloading 100s of MBs just to throw away is huge overhead to have here for ~very little benefit I can see.

gfs (changes requested):

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: This was in response to https://github.com/microsoft/OSSGadget/pull/329#discussion_r853207419
I could change it to a request for just the headers to see if we 404 or not. We just want to see if the uri is even valid.

src/Shared/PackageManagers/NPMProjectManager.cs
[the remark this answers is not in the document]
gfs: Strongly disagree. At least in the current implementation the check is to download the entire artifact which is of arbitrary size adding huge overhead to this call that it doesn't need. 

The contract is to provide what the package repository has on the package. If the URI can be accessed or not at the moment that it is being enumerated isn't relevant. 

In fact, it seems to me that not returning Uris that are listed breaks the contract and creates a disconnect between the data you can see in the repository and what the api returns in a way that is invisible to the consumer of the API.

Finally the consumer then would be led to believe that the repository doesn't have information about the package - and they'd lose the information about that particular artifact being listed but not accessible at that moment for whatever reason. 

Please revert this change.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: I could change it to:
` HttpResponseMessage result = await httpClient.GetAsync(artifactUri, HttpCompletionOption.ResponseHeadersRead);`
https://docs.microsoft.com/en-us/dotnet/api/System.Net.Http.HttpCompletionOption?view=net-6.0#fields

gfs: I added a more in depth comment to that as well. I don't believe we should even be doing that.

However, if you believe it is needed, and it is possible to reduce the network calls to be of minimal size, you will need to change artifact uri and still return those Uris that are inaccessible but have a field in the artifact uri for the status code you got for the headers. 

I understand why a user would want this logic. It just doesn't belong here.

jpinz: We've used that in CFS before. This allows us to get a response to see if a given URL 404s, without waiting to retrieve the content. I'm also ok with this being up to the consumer, where they have to check if the uri being returned exists or not. 
We could also add a parameter to the method to check if the uri exists, and set a default to true or false depending on what you think.

jpinz: I'm now leaning towards removing it and leaving it up to the consumer to implement. If for whatever reason the call fails, we don't want to falsely claim the artifact doesn't exist. And I don't think it's OSS Gadget's job to add the retry logic to assert that the artifact does or doesn't exist.

gfs: Retry logic is another good point.  This adds a lot of uncertainty to this api.

Another example, I think it is theoretically possible you could have a repository where to access packages requires a secret in the headers with different permissions than to read.

It seems to be to be preferable for the failure to come when the consumer of the api tries to get the package and can perform their own retry logic as appropriate rather than to pretend the uri doesn't exist.

jpinz: Yeah, private package registries/repositories would mess with all of this.

jpinz: I reverted it. Resolved.

src/Shared/PackageManagers/NPMProjectManager.cs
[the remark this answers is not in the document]
gfs: I think the alternative might that you want a different call that does, with some retry logic, determine if the package that this api returns exists.

I'm fine with adding that as a separate method, using the header shortcut you mentioned in the other thread.

Or perhaps it is an extension method to HttpClient?

```c#
bool UriExists(this HttpClient client, Uri uri, RetryOptions opts)
{
...
}
```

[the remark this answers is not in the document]
jpinz: I decided to forgo adding this to OSS Gadget. It's simple enough to implement on the consumer side. I did add a remark to the method clarifying that this just returns the URI(s) we expect to exist, not confirming that they exist.
@bhuwan-agarwal we might have to re-evaluate this for `MavenProjectManager` as we can't "assume" certain files exist or not unlike NPM, NuGet, and PyPI where we know the exact file associated with the given `purl`.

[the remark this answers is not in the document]
bhuwan-agarwal: @gfs, a separate method makes sense. That way the caller can chose to do a check if they want to before downloading it.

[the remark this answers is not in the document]
jpinz: I could write a method in `BaseArtifactUri` that takes in an `HttpClient` and optionally a Polly `Policy` that does the check. That way the consumer can use it if they would like to.

src/Shared/Model/BaseArtifactUri.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// </summary>
+    public string Extension { get; } = Extension;
+
+    /// <summary>
+    /// Check to see if the <see cref="Uri"/> exists.
+    /// </summary>
+    /// <param name="httpClient">The <see cref="HttpClient"/> to use when checking if the <see cref="Uri"/> exists.</param>
+    /// <param name="policy">An optional <see cref="AsyncRetryPolicy"/> to use with the http request.</param>
+    /// <returns>If the request returns <see cref="HttpStatusCode.OK"/>.</returns>
+    public async Task<bool> ExistsAsync(HttpClient httpClient, AsyncRetryPolicy<HttpResponseMessage>? policy = null)
```
bhuwan-agarwal, line 37: Should there be a test for this method?

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -37,6 +38,17 @@ public NPMProjectManager(string destinationDirectory) : base(destinationDirector
         {
         }
 
+        /// <inheritdoc />
+        public override IEnumerable<NPMArtifactUri> GetArtifactDownloadUris(PackageURL purl)
```
gfs, line 42: Are there other artifacts we want to populate for npm? Package.json maybe?

src/Shared/PackageManagers/PyPIProjectManager.cs
```diff
@@ -429,5 +436,34 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
+            /// <param name="uri">The string of the uri this artifact can be found at.</param>
+            /// <param name="extension">The file extension for the file found at the <paramref name="uri"/>.</param>
+            public PyPIArtifactUri(PyPIArtifactType type, string uri, string? extension = null) : this(type, new Uri(uri), extension) { }
+        }
+
+        public enum PyPIArtifactType
+        {
+            Unknown = 0,
+            Tarball,
+            Wheel,
```
gfs, line 466: I don't see that you produce artifacts with this value.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -342,5 +352,36 @@ private static string GetNupkgUrl(string id, string version)
...
+            /// <param name="extension">The file extension for the file found at the <paramref name="uri"/>.</param>
+            public NuGetArtifactUri(NuGetArtifactType type, string uri, string? extension = null) : this(type, new Uri(uri), extension) { }
+        }
+
+        public enum NuGetArtifactType
+        {
+            Unknown = 0,
+            Binary,
+            Nupkg,
+            Nuspec,
```
gfs, line 384: Are you going to add these to the enumeration above in this PR?

gfs (commented):
What are the plans for enumerating the other types of artifacts that are defined in this PR?

src/Shared/Model/BaseArtifactUri.cs
jpinz: There is, in the ProjectManagerTests.

src/Shared/PackageManagers/PyPIProjectManager.cs
jpinz: I had written it in preparation for eventually using it, but yeah I wasn't planning on using it in this PR. Older PyPI packages didn't include a `.whl` file so I wasn't sure if I wanted to construct them without confidence that it would exist.

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -101,15 +101,15 @@ public ProjectManagerFactory(IHttpClientFactory? httpClientFactory = null)
...
+                        new NPMProjectManager(destinationDirectory, new NoOpPackageActions(), httpClientFactory)
                 },
                 {
                     NuGetProjectManager.Type, destinationDirectory =>
                         new NuGetProjectManager(destinationDirectory, new NuGetPackageActions(), httpClientFactory) // Add the NuGetPackageActions to the NuGetProjectManager.
                 },
                 {
                     PyPIProjectManager.Type, destinationDirectory =>
-                        new PyPIProjectManager(httpClientFactory, destinationDirectory)
+                        new PyPIProjectManager(destinationDirectory, new NoOpPackageActions(), httpClientFactory)
```
pmalmsten, line 112: I'm curious what the motivation of using `NoOpPackageActions` here is?

src/Shared/Model/ArtifactUri.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    /// <summary>
+    /// The <see cref="Uri"/> for where this artifact can be found online.
+    /// </summary>
+    public Uri Uri { get; }
+
+    /// <summary>
+    /// The file extension for this artifact file.
+    /// </summary>
+    public string Extension { get; }
```
pmalmsten, line 56: We might want to be careful with the notion of 'file extension' here. For example, it seems possible that one or more package managers might choose to represent artifacts by non-HTTP URIs that a client library would know how to fetch.

As an alternative, what if this property was actually a function that attempts to extract a file name from the `Uri` field on this object? If the given `Uri` does not represent a file, it could return null or throw an exception?

src/Shared/Model/ArtifactUri.cs
pmalmsten: The revised `ArtifactUri` approach looks great - thanks for making the changes! Resolving this thread.

src/Shared/PackageManagers/PyPIProjectManager.cs
jpinz: I can't find a way to construct the `.whl` file uris. I put a TODO down to figure it out.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: I removed `Binary` but I am doing `Nuspec` now too.

src/Shared/PackageManagers/ProjectManagerFactory.cs
jpinz: I mean I could just use null instead. But the idea is that I don't need a PackageActions implementation on NPM or PyPI right now, as all of the methods are already defined in the ProjectManager.
I can refactor it if we want to, but a NoOpPackageActions was just easier.

src/Shared/Model/ArtifactUri.cs
jpinz: The latter is basically what we did right now anyways. Nowhere was I actually using the `extension` constructor argument, it was always being constructed by `Path.GetExtension(uri.AbsolutePath)`.
How's that seem?

src/Shared/Model/ArtifactUri.cs
```diff
@@ -2,46 +2,84 @@
...
-        // PyPI specific
-        Wheel,
-
-        // Maven specific
-        Pom,
-        Jar,
-        Sources,
+        policy ??= DefaultPolicy;
+
+        return (await policy.ExecuteAsync(() => httpClient.GetAsync(Uri, HttpCompletionOption.ResponseHeadersRead))).StatusCode == HttpStatusCode.OK;
```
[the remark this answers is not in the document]
jpinz: I moved it to `TypedManager.UriExistsAsync`, I think that makes more sense. It takes in a `Uri` and the optional `AsyncRetryPolicy<HttpResponseMessage>`

src/Shared/Model/ArtifactUri.cs
```diff
@@ -2,46 +2,84 @@
...
+        Backoff.DecorrelatedJitterBackoffV2(TimeSpan.FromMilliseconds(100), retryCount: 5);
+
+    /// <summary>
+    /// The default policy to use when checking to see existence of the <see cref="Uri"/>.
+    /// </summary>
+    private static readonly AsyncRetryPolicy<HttpResponseMessage> DefaultPolicy =
+        Policy
+            .Handle<HttpRequestException>()
+            .OrResult<HttpResponseMessage>(r => r.StatusCode != HttpStatusCode.OK) // Also consider any response that doesn't have a 200 status code to be a failure.
+            .WaitAndRetryAsync(Delay);
```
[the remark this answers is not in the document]
jpinz: Same as above.

bhuwan-agarwal (approved):
Looks good from my perspective.

src/Shared/Model/ArtifactUri.cs
```diff
@@ -0,0 +1,85 @@
...
+/// <typeparam name="T">The enum to represent the artifact type.</typeparam>
+public record ArtifactUri<T> where T : Enum
+{
+    /// <summary>
+    /// Initializes a new instance of <see cref="ArtifactUri{T}"/>.
+    /// </summary>
+    /// <param name="type">The type of artifact for this <see cref="ArtifactUri{T}"/>.</param>
+    /// <param name="uri">The <see cref="Uri"/> this artifact can be found at.</param>
+    /// <param name="extension">The file extension for the file found at the <paramref name="uri"/>.</param>
+    public ArtifactUri(T type, Uri uri, string? extension = null)
```
gfs: Why is extension configurable separate from uri? For .tar.gz?

src/Shared/Model/ArtifactUri.cs
gfs, line 54:
```suggestion
    public string Extension => Path.GetExtension(uri.AbsolutePath);
```
I think its better if these cannot be out of sync

src/Shared/PackageActions/NoOpPackageActions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.CST.OpenSource.PackageActions;
+
+using Contracts;
+using PackageUrl;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+public class NoOpPackageActions : IManagerPackageActions<IManagerPackageVersionMetadata>
```
gfs, line 11: Is this really no-op if it returns not implemented? I think either the operations should just return empty results or it should be renamed NotImplementedActions.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -18,7 +19,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
     using System.Text.Json;
     using System.Threading.Tasks;
 
-    public class NuGetProjectManager : TypedManager<NuGetPackageVersionMetadata>
+    public class NuGetProjectManager : TypedManager<NuGetPackageVersionMetadata, NuGetProjectManager.NuGetArtifactType>
```
gfs, line 22: Is there a case where we would have non-matching package metadata and enum type here? should the enum be defined in the meta data class and then referenced by the manager so it doesn't require two type params?

gfs (commented):
I don't think I see anything that blocks merging if the consensus is this is good but I have some comments/questions.

src/Shared/Model/ArtifactUri.cs
jpinz: This is removed from the most recent commit. Looks like you reviewed an older version.

jpinz: But to answer your question, it wasn't ever being used so it doesn't really matter. But yeah there is the issue with the .tar.gz files still, so if you know of a good solution please let me know!

src/Shared/PackageActions/NoOpPackageActions.cs
jpinz: This is a half measure until a larger refactor of ProjectManagers is done with metadata and whatnot, but I agree. I'll just have them return empty results instead of not implemented. That was just the default return when I generated it.

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: I could merge them as a part of the larger metadata refactoring work, which I just started working on today.

src/Shared/Model/ArtifactUri.cs
gfs: Some untested thoughts:
1. You could take everything after the last '/' and then find the first the first '.' in that and take everything after that '.' as an extension.
2. You could use the Get extension thing below and just have special cases for .tar.* (also possible to have like a .tar.bz2, we shouldn't just restrict to tar.gz.

jpinz: The first one wouldn't work because of file names like this: https://pypi.org/packages/source/r/requests/requests-2.27.1.tar.gz 
So I'll mess around with the second one

jpinz: Had to write my own method to do it. Just pushed.

gfs: I think that looks like a fine way to do it.

src/Shared/PackageManagers/NuGetProjectManager.cs
gfs: I think that makes sense to do before converting everything to typed managers. I cannot think of a case where you'd want ones that don't match very closely.

gfs (approved):

src/Shared/PackageManagers/NuGetProjectManager.cs
jpinz: Tracking this work with #333

Issue #331, opened 2022-04-16
Wildcard downloads broken for NuGet


scovetta:
I noticed that NuGet wildcard downloads are broken, see screenshot below.

![image](https://user-images.githubusercontent.com/732166/163694515-c8b77798-b35c-4dc1-b37a-728980f0e08e.png)

This is using version 0.1.332 on Windows.

I'm expecting the wildcard version to mean "download all available versions".

Similar for another NuGet project.
![image](https://user-images.githubusercontent.com/732166/163694535-63ae958a-0708-454a-92bf-bfbf0c269dfd.png)

This works for npm:
![image](https://user-images.githubusercontent.com/732166/163694546-9d3882bc-ba5e-4b6c-b439-4e1c54405842.png)

scovetta:
FYI, this works properly on 0.1.312, it's probably related to the refactoring to use the NuGet library instead of going direct.

![image](https://user-images.githubusercontent.com/732166/163694989-df0ebcdf-b166-455a-825c-f525e1afc8df.png)

jpinz:
I'll look into it now!

jpinz:
@scovetta Hopefully should be fixed now. Let me know if you run into any other issues.

PR #332, opened 2022-04-18
Fix wildcard version NuGet downloads.

jpinz:
This PR fixes: #331

### DIFF
--- a/src/Shared/PackageActions/NuGetPackageActions.cs
+++ b/src/Shared/PackageActions/NuGetPackageActions.cs
@@ -67,7 +67,7 @@ public class NuGetPackageActions : IManagerPackageActions<NuGetPackageVersionMet
         // If we want to extract the contents of the .nupkg, send it to ArchiveHelper.ExtractArchiveAsync.
         if (doExtract)
         {
-            return await ArchiveHelper.ExtractArchiveAsync(topLevelDirectory, Path.GetExtension(targetPath) + ".nupkg", packageStream, cached);
+            return await ArchiveHelper.ExtractArchiveAsync(topLevelDirectory, targetPath + ".nupkg", packageStream, cached);
         }
 
         return filePath;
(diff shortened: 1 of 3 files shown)


jpinz:
Writing some tests for this case now.

jpinz:
Fixing tests now, I also realized it was not a good test because as soon as a new version of any of the ones I wrote tests for would break.

pmalmsten (approved):

gfs (commented):
I don't see a code change in here that handles .* versions differently. Can you explain what is fixed?

jpinz:
> I don't see a code change in here that handles .* versions differently. Can you explain what is fixed?

When debugging I realized it wasn't an issue with handling * versions, but it was an issue with the `TypedManager` which only gets used by NuGet. It was trying to enumerate through extracted files every time, even if we weren't extracting the `.nupkg` file. There might be more that I'm forgetting, so I'll go debug and make sure that was all.
Also I found the issue with the `NuGetPackageActions` where when extracting the file it was the extension of the file + `.nupkg` which is incorrect.

jpinz:
> > I don't see a code change in here that handles .* versions differently. Can you explain what is fixed?
> 
> When debugging I realized it wasn't an issue with handling * versions, but it was an issue with the `TypedManager` which only gets used by NuGet. It was trying to enumerate through extracted files every time, even if we weren't extracting the `.nupkg` file. There might be more that I'm forgetting, so I'll go debug and make sure that was all. Also I found the issue with the `NuGetPackageActions` where when extracting the file it was the extension of the file + `.nupkg` which is incorrect.

Just confirmed it yeah, the `PackageDownloader` constructor did successfully add every version to the `PackageVersions = List<PackageURL>` but the `TypedManager.DownloadVersionAsync` is where it was failing because we were trying to assert that we had extracted **every single** artifact we downloaded, even when we didn't specify that we wanted to extract the file we downloaded.

PR #334, opened 2022-04-26
Add GetPublishedAtAsync to BaseProjectManager

jpinz:
It returns the DateTime of a given purl (with a specified version) for when it was published. Or returns null if it couldn't be found.
For CFS we are using this data, so having a direct method to get it vs calling GetPackageMetadata and only keeping the time would help. If you disagree no worries and we can just do that, this was just a fairly easy thing to whip up.

Also, `NPMProjectManager` wasn't setting `UploadTime` in the `GetPackageMetadata` call.

### DIFF
--- a/src/Shared/PackageManagers/BaseProjectManager.cs
+++ b/src/Shared/PackageManagers/BaseProjectManager.cs
@@ -291,6 +291,17 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         {
             throw new NotImplementedException("BaseProjectManager does not implement EnumerateVersions.");
         }
+        
+        /// <summary>
+        /// Gets the <see cref="DateTime"/> a package version was published at.
+        /// </summary>
+        /// <param name="purl">Package URL specifying the package. Version is mandatory.</param>
+        /// <param name="useCache">If the cache should be used when looking for the published time.</param>
+        /// <returns>The <see cref="DateTime"/> when this version was published, or null if not found.</returns>
+        public virtual Task<DateTime?> GetPublishedAtAsync(PackageURL purl, bool useCache = true)
+        {
+            throw new NotImplementedException("BaseProjectManager does not implement GetPublishedAtAsync.");
+        }
 
         /// <summary>
         /// Gets the latest version from the package metadata.
--- a/src/Shared/PackageManagers/BaseProjectManager.cs
+++ b/src/Shared/PackageManagers/BaseProjectManager.cs
@@ -291,17 +291,6 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         {
             throw new NotImplementedException("BaseProjectManager does not implement EnumerateVersions.");
         }
-        
-        /// <summary>
-        /// Gets the <see cref="DateTime"/> a package version was published at.
-        /// </summary>
-        /// <param name="purl">Package URL specifying the package. Version is mandatory.</param>
-        /// <param name="useCache">If the cache should be used when looking for the published time.</param>
-        /// <returns>The <see cref="DateTime"/> when this version was published, or null if not found.</returns>
-        public virtual Task<DateTime?> GetPublishedAtAsync(PackageURL purl, bool useCache = true)
-        {
-            throw new NotImplementedException("BaseProjectManager does not implement GetPublishedAtAsync.");
-        }
 
         /// <summary>
         /// Gets the latest version from the package metadata.
--- a/src/Shared/PackageManagers/NPMProjectManager.cs
+++ b/src/Shared/PackageManagers/NPMProjectManager.cs
@@ -179,8 +179,8 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         public async Task<DateTime?> GetPublishedAtAsync(PackageURL purl, bool useCache = true)
         {
             Check.NotNull(nameof(purl.Version), purl.Version);
-            string? uploadTime = (await this.GetPackageMetadataAsync(purl, useCache))?.UploadTime;
-            return uploadTime == null ? null : DateTime.Parse(uploadTime);
+            DateTime? uploadTime = (await this.GetPackageMetadataAsync(purl, useCache))?.UploadTime;
+            return uploadTime;
         }
 
         /// <summary>
@@ -261,7 +261,10 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                 if (root.TryGetProperty("time", out JsonElement time))
                 {
                     string? uploadTime = OssUtilities.GetJSONPropertyStringIfExists(time, metadata.PackageVersion);
-                    metadata.UploadTime = uploadTime;
+                    if (uploadTime != null)
+                    {
+                        metadata.UploadTime = DateTime.Parse(uploadTime);
+                    }
                 }
 
                 if (versionElement != null)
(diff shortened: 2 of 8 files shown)


mounika1837 (approved):

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -291,6 +291,17 @@ public virtual Task<IEnumerable<string>> EnumerateVersionsAsync(PackageURL purl,
...
+        /// <summary>
+        /// Gets the <see cref="DateTime"/> a package version was published at.
+        /// </summary>
+        /// <param name="purl">Package URL specifying the package. Version is mandatory.</param>
+        /// <param name="useCache">If the cache should be used when looking for the published time.</param>
+        /// <returns>The <see cref="DateTime"/> when this version was published, or null if not found.</returns>
+        public virtual Task<DateTime?> GetPublishedAtAsync(PackageURL purl, bool useCache = true)
+        {
+            throw new NotImplementedException("BaseProjectManager does not implement GetPublishedAtAsync.");
+        }
```
pmalmsten, line 304: Nit: I wonder whether it would make sense to hold off on adding new methods like this to `BaseProjectManager` until there is 'critical mass' per se of implementations for them for the specific package managers.

Advertising a method like this for all package managers when it only actually works for a small subset is not a great developer experience to have by default.

jpinz: I get what you're saying. As far as I see it then we have two options. We have this one. And the other one is the consumer has to cast the `BaseProjectManager` to one of the 3 that implement the method then they can call it. Personally I would rather do it this way, but what do you think?

pmalmsten: CFS is the only consumer of this method today right? And if I recall correctly, we are planning to instantiate the correct type of project manager directly anyway?

If that's true, then `BaseProjectManager` only really serves callers who want to work with more than one type of package at a time / who do not know which type of project manager instance they are getting. If we don't really have any consumers of this data yet from that perspective, I wonder if that gives us time to hold off on adding it to `BaseProjectManager` until that use case either materializes, or when we finish implementing this for many/most package managers.

jpinz: Yeah that's a good point. I will have it in just the ones we need it in.

gfs: Is the published time available in the metadata from the metadata method? The default behavior could be to call the metadata method and then return the time.

gfs: Ah I see metadata method isn't standard in the format it returns. That seems to be an opportunity for a metadata class with basic things like the requested purl, publish time etc. which would enable a default implementation of the publish time by calling the base metadata call - overrideable if there is a more efficient way for that specific manager.

gfs (commented):
Per comments hold off on the base project manager implementation unless there is good coverage. But keep the methods such that it can be added later to the base. 

I think it would make sense to change what is returned from base managers metadata call to be a parsed metadata object rather than the raw string - though the string itself could also be included as a field in the metadata object.

pmalmsten:
> I think it would make sense to change what is returned from base managers metadata call to be a parsed metadata object rather than the raw string - though the string itself could also be included as a field in the metadata object.

+1 to this (though it sounds too big for this PR of course). When callers are writing code against the `BaseProjectManager` interface, they are most likely trying to work with one or more different types of packages as uniformly as possible, so anything we put on `BaseProjectManager` should emphasize that desired uniformity (i.e. provide parsed metadata objects as return values, methods should function for all / as many as possible different package managers, etc.)

pmalmsten (approved):

jpinz:
> I think it would make sense to change what is returned from base managers metadata call to be a parsed metadata object rather than the raw string - though the string itself could also be included as a field in the metadata object.

I've been working on something very close to this, meeting with @bhuwan-agarwal tomorrow to discuss it further.

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: Realistically we could just do something like this on the `BaseProjectManager` in this method. And then it relies on `GetPackageMetadataAsync` being implemented but otherwise works as expected:
``` cs
PackageMetadata? metadata = await this.GetPackageMetadataAsync(purl, useCache);
string? uploadTime = metadata?.UploadTime;
if (uploadTime == null)
{
  return null;
}
return DateTime.Parse(uploadTime);

File: src/Shared/PackageManagers/NPMProjectManager.cs [hunk 2]
```diff
@@ -170,6 +170,36 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
...
+        /// </summary>
+        /// <param name="purl">Package URL specifying the package. Version is mandatory.</param>
+        /// <param name="useCache">If the cache should be used when looking for the published time.</param>
+        /// <returns>The <see cref="DateTime"/> when this version was published, or null if not found.</returns>
+        public async Task<DateTime?> GetPublishedAtAsync(PackageURL purl, bool useCache = true)
+        {
+            string? content = await GetMetadataAsync(purl, useCache);
+            if (string.IsNullOrEmpty(content)) { return null; }
+
+            // convert NPM package data to normalized form
```
gfs, line 184: This comment doesn't seem correct to me for the next few lines

gfs (approved):

src/Shared/PackageManagers/BaseProjectManager.cs
pmalmsten: For how many package managers is the `metadata?.UploadTime` value non-null? If many/most provide a value, then go for it. If only a handful do, then I'm not sure this would be all that better than throwing `NotImplementedException` as of now.

jpinz: Only the 3 do right now. I'll just do this, but on the Manager implementation side vs doing all the parsing. It simplifies it a lot.

pmalmsten: (Also, double checking an assumption - `DateTime.Parse` will be able to understand the string value in `metadata?.UploadTime` for all package managers right?)

jpinz: Yes, and I confirmed it in the unit tests.

gfs: It seems to me the upload time value in the metadata object should already be a DateTime.

jpinz: I could change it to parse it elsewhere that makes sense. I'll go do it real quick.

gfs: What about having this in the base, once the metadata is returned as an object?
```c#
public async DateTime GetPublishTimeAsync(PackagURL Purl) => await GetMetaDataAsync(Purl)?.PublishTime
```

This requires having a metadata object be returned
```c#
public class Metadata
{
    public DateTime PublishTime {get; set;}
    public string Raw { get; set }
}
```

jpinz: Yeah that can be something we look into when we do the larger rework to Metadata.

gfs (approved):

pmalmsten (approved):

PR #339, opened 2022-06-24
Fix broken formatting for oss-reproducible.

scovetta:
I missed a $-prefix on a few output lines, so the literal `{filesString}` was being shown instead of the content.


gfs (approved):

scovetta:
Merged via #338, closing this one.

PR #341, opened 2022-06-29
Initial import of Codespaces config.

scovetta:
This PR adds a few files which helps developers work on OSS Gadget via Codespaces.


gfs (approved):
Seems fine. I don't know enough about codespaces to validate if it's correct, but I don't see any issues.

PR #345, opened 2022-08-22
Cargo package existence and package version existence logic update.

jpinz:
The check for if a package exists, or if a package version exists, and the list of package versions that do exist for a package is a bit broken for cargo, it appears to be constantly getting 403s as we get throttled. 

This PR updates it to use their registry mirror on github as per the recommendations on https://crates.io/data-access to use https://github.com/rust-lang/crates.io-index


scovetta (approved):

jpinz:
@scovetta looks like the libraries.io unit test is failing on the pipeline? I got it to work locally after pasting in my API Key, is it not set up correctly for the pipeline?

scovetta:
I don't think they expire, might be getting throttled. I'll take a look.

src/Shared/Model/Enums/JsonParsingOption.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+
+namespace Microsoft.CST.OpenSource.Model.Enums;
+
+/// <summary>
+/// Special cases used for json parsing.
+/// </summary>
+public enum JsonParsingOption
+{
+    Unknown = 0,
```
brbayes-msft, line 10: Maybe do None instead of Unknown here since this isn't a value that will be serialized across the wire.

brbayes-msft (approved):

src/Shared/Model/Enums/JsonParsingOption.cs
jpinz: Can do!

gfs (approved):

src/Shared/Model/Enums/JsonParsingOption.cs
jpinz: Switched it from unknown to none, and made it so that instead of the option being nullable in `BaseProjectManager` to using `None` instead.

brbayes-msft (approved):

PR #349, opened 2022-09-06
Add some NPM methods to support checking if a version was unpublished, or if a package is a security holding pacakge.

jpinz:
We ran into an issue where we had marked a package as existing and then ingested it, but upon checking it again we discovered that it had since been pulled from the NPM registry. So we needed a way to check if an NPM package (only doing NPM for now) existed but has since been unpublished.
Similarly, we wanted a way to check if a package was a security holding package. For example: https://www.npmjs.com/package/http


src/oss-tests/ProjectManagerTests/NPMProjectManagerTests.cs
```diff
@@ -111,6 +114,41 @@ public async Task PackageVersionDoesntExistsAsyncSucceeds(string purlString)
...
+        [DataTestMethod]
+        [DataRow("pkg:npm/%40somosme/webflowutils@1.0.0")]
+        [DataRow("pkg:npm/%40somosme/webflowutils@1.2.3", false)]
+        public async Task PackageVersionPulledAsync(string purlString, bool shouldSucceed = true)
+        {
+            PackageURL purl = new(purlString);
+
+            if (shouldSucceed)
+            {
+                Assert.IsTrue(await _projectManager.PackageVersionPulled(purl, useCache: false));
```
gfs, line 126:
```suggestion
                Assert.AreEqual(shouldSucceed, await _projectManager.PackageVersionPulled(purl, useCache: false));
```

Nit: This is simpler by using the bool argument with assert.areequal - replacing the entire if + else. Also would then make sense to change the input variable name to `expected`.

src/oss-tests/ProjectManagerTests/NPMProjectManagerTests.cs
```diff
@@ -111,6 +114,41 @@ public async Task PackageVersionDoesntExistsAsyncSucceeds(string purlString)
...
+        [DataRow("pkg:npm/lodash.js")]
+        [DataRow("pkg:npm/lodash.js@1.0.0")]
+        [DataRow("pkg:npm/lodash", false)]
+        public async Task PackageSecurityHoldingAsync(string purlString, bool shouldHaveSecurityHolding = true)
+        {
+            PackageURL purl = new(purlString);
+
+            if (shouldHaveSecurityHolding)
+            {
+                Assert.IsTrue(await _projectManager.PackageSecurityHolding(purl, useCache: false));
```
gfs, line 144: Same as earlier comment.

gfs (approved):
Looks fine. Minor recommendations to make the tests a bit simpler.

src/oss-tests/ProjectManagerTests/NPMProjectManagerTests.cs
jpinz: Done!

PR #352, opened 2022-09-09
Add new methods to help with determining if and why a package was removed from the registry.

jpinz:
Also adds a method `EnumerateVersionsWithPublishTimeAsync` for getting a dictionary with each version and when they were published. (Only implemented for NPM as of now)


src/Shared/Model/PackageExistence.cs
```diff
@@ -0,0 +1,65 @@
...
+        this.PackageDidExist = packageDidExist;
+        this.DoesNotExistReasons = doesNotExistReasons ?? new List<PackageDoesNotExistReason>();
+    }
+
+    public IEnumerable<PackageDoesNotExistReason> DoesNotExistReasons { get; init; }
+    public bool DidExist { get; init; }
+    public bool Exists { get; init; }
+    public bool PackageDidExist { get; init; }
+    public bool PackageExists { get; init; }
+}
```
pmalmsten, line 65: This feels a bit complex to me. Here's an alternative which looks a bit simpler to my eyes (the below is just for package version - we would do the same for package as well:

```c#

public enum PackageVersionRemovalReason
{
    Unpublished,
    ConsideredMalicious,
}

interface IPackageVersionExistence
{
    bool Exists { get; }
    bool HasEverExisted { get; }
    bool WasRemoved => !this.Exists && this.HasEverExisted;
}

public record PackageVersionExists : IPackageVersionExistence
{
    public bool Exists => true;
    public bool HasEverExisted => true;
}

public record PackageVersionNotFound : IPackageVersionExistence
{
    public bool Exists => false;
    public bool HasEverExisted => false;
}

public record PackageVersionRemoved(IReadOnlySet<PackageVersionRemovalReason> RemovalReasons) : IPackageVersionExistence
{
    public bool Exists => false;
    public bool HasEverExisted => true;
}
```

What do you think? 

One of the advantages of the above format is that it lends itself to [pattern matching by type](https://endjin.com/blog/2022/02/pattern-matching-in-csharp) rather nicely.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -310,6 +310,19 @@ public virtual Task<IEnumerable<string>> EnumerateVersionsAsync(PackageURL purl,
...
+        
+        /// <summary>
+        /// Enumerates all possible versions of the package identified by purl, in descending order with the time they were published at.
+        /// </summary>
+        /// <remarks>The latest version is always first, then it is sorted by SemVer in descending order.</remarks>
+        /// <param name="purl">Package URL specifying the package. Version is ignored.</param>
+        /// <param name="useCache">If the cache should be used when looking for the versions.</param>
+        /// <param name="includePrerelease">If pre-release versions should be included.</param>
+        /// <returns> A dictionary where the key is the package version, and the value is the time it was published at.</returns>
+        public virtual Task<IDictionary<string, DateTime>> EnumerateVersionsWithPublishTimeAsync(PackageURL purl, bool useCache = true, bool includePrerelease = true)
```
pmalmsten, line 322: This kind of interface feels a bit narrow / not very extensible. Are there any cases where we would get a performance benefit from returning a version-to-datetime mapping, as opposed to a version-to-metadata-object mapping, where the metadata object can include datetime among other things?

I ask because it would not be a great user experience to have a bunch of separate version-to-something mapping methods, if/when we wanted to return more information like this.

pmalmsten, line 322: Do we have any package-type-agnostic use cases in mind for this method at this point? Or could this just be defined on the NPM-specific project manager to start?

We don't want users of the library to frequently stumble on `NotImplementedException` when we haven't implemented base methods like this for many package types yet.

src/Shared/Model/Enums/PackageDoesNotExistReason.cs
```diff
@@ -0,0 +1,24 @@
...
+    
+    /// <summary>
+    /// When the package was unpublished from the repository/registry by someone.
+    /// </summary>
+    PackageUnpublished,
+
+    /// <summary>
+    /// When the package/version was removed from the repository/registry by the package manager's administrators due to a security risk.
+    /// </summary>
+    RemovedForSecurity,
```
pmalmsten, line 23: `RemovedForSecurity` is pretty vague. Can we be more specific?

Maybe something like `ConsideredMalicious`? That would emphasize 'this looks like malware' as opposed to just has a known vulnerability/CVE.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -364,14 +459,36 @@ public virtual async Task<bool> PackageVersionExistsAsync(PackageURL purl, bool
...
+        /// <summary>
+        /// Check to see if the package was removed for security.
+        /// </summary>
+        /// <param name="purl">The <see cref="PackageURL"/> to check.</param>
+        /// <param name="useCache">If the cache should be checked for the existence of this package.</param>
+        /// <returns>True if this package is was removed for security. False otherwise.</returns>
+        public virtual Task<bool> PackageRemovedForSecurity(PackageURL purl, bool useCache = true)
+        {
+            return Task.FromResult(false);
         }
```
pmalmsten, line 492: I see what you were going for by having these be helper methods for the above package existence methods, but would anyone actually call these methods? They would probably call the package existence one right, because that one provides any/all such reasons at once?

If callers will probably not use these helper methods, should these just be implementation details on the type-specific managers? I know we would have to move the implementation of the package existence method to the type-specific managers, but that probably makes sense anyway - having these separate method calls do separate JSON parsing does not seem very efficient.

pmalmsten (changes requested):

gfs (approved):
Looks fine I think. Am trusting your test cases address appropriately.

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: This could be my lack of knowledge, but I would imagine that there is a _slight_ efficiency increase if we are just trying to get a version-to-datetime mapping vs a version-to-metadata then selecting just the datetime from the metadata. But I could also see how there is such a negligible efficiency increase (if any at all) to the point where it doesn't really matter much anyways.

jpinz: Right now it is just npm yes, I just didn't know if it made more sense from our side of things to handle a `NotImplementedException` vs doing a check for it being an NPM package, and then casting the `BaseProjectManager` to `NpmProjectManager` to then call the method.

src/Shared/Model/Enums/PackageDoesNotExistReason.cs
jpinz: This is just trying to convey when the registry itself pulls a package/version from the registry. I'd imagine that the only use case there is for "Malicious" purposes, so I think `ConsideredMalicious` makes sense, but if anyone else has another suggestion let me know!

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: Yeah I could have them be implementation specific that makes sense. I don't see a use case for calling this method specifically when the new existence check + record exists.

src/Shared/Model/PackageExistence.cs
```diff
@@ -0,0 +1,65 @@
...
+            throw new InvalidOperationException("The package can't have Exists as true, and DidExist as false!");
+        }
+
+        this.Exists = exists;
+        this.DidExist = didExist;
+        this.DoesNotExistReasons = doesNotExistReasons ?? new List<PackageDoesNotExistReason>();
+
+        if (this.DoesNotExistReasons.Contains(PackageDoesNotExistReason.VersionUnpublished))
+        {
+            throw new InvalidOperationException("The package can't have VersionUnpublished as a does not exist reason!");
```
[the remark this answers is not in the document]
jpinz: I removed this logic so no longer a problem!

src/Shared/Model/PackageExistence.cs
jpinz: Addressed!

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: Addressed!

src/Shared/PackageManagers/BaseProjectManager.cs
gfs: I think any real improvement you’d get would be if the repository you’re talking to has an api specifically for this. If you have to fetch and decode the whole metadata and then just return the time stamp that’s not really saving anything over returning the whole parsed metadata - assuming the metadata with time is fetched in the same network call. 

That said, that now means that you need to implement that for every manager to get the benefit, which is even in the best case going to be slight I believe. 

The rub is, however, that we don’t have a standard metadata format (I believe?) so there isn’t a different place to slot in the DateTime. I do however think that it would be better to just return the metadata and let the caller look at the fields they want. a caller would then be able to just `.Select(x=> (x.version, x.datetime))` on the returned metadata if they wanted this behavior.

gfs: @jpinz if you don’t intend to implement this for other managers it does not belong on baseprojectmanager.

src/Shared/Model/Enums/PackageDoesNotExistReason.cs
gfs: I think both `RemovedForSecurity` and `ConsideredMalicious` convey a sense of confidence in the rationale for removal we don’t truly have. I recommend `RemovedByRepository`

gfs (changes requested):
Reading more of the discussions it’s not clear to me that these methods belong on base project manager if they won’t be implemented broadly, and perhaps some of them are just too specific at all for ossgadget and would be better as extension methods in your project.

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: I'm removing it from this PR until we can come up with a good solution for getting metadata for all versions of a given package.

jpinz: There is a standard metadata format `PackageMetadata` which includes `UploadTime`, the bigger issue is that right now the metadata request only ever gets the values for a single version, we don't have a method for getting say a dictionary where the key is the version, and the value is the version specific metadata. Which would remove the need for this method and we would just replace it with something along the lines of the `.Select` you mentioned above.

jpinz:
> Reading more of the discussions it’s not clear to me that these methods belong on base project manager if they won’t be implemented broadly, and perhaps some of them are just too specific at all for ossgadget and would be better as extension methods in your project.

I removed the `EnumerateVersionsWithPublishTimeAsync` method until we can think of a better solution and just kept the logic for the unpublishing work which is what the main goal of this PR was for.

src/Shared/PackageManagers/BaseProjectManager.cs
gfs: Ah I see. I think we can think of a way to structure that object so this data is in it.

src/Shared/Contracts/IPackageExistence.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Gets if the package/version was removed.
+    /// So if it doesn't exist now, but at one point did exist.
+    /// </summary>
+    bool WasRemoved => !this.Exists && this.HasEverExisted;
+
+    /// <summary>
+    /// Gets the reasons (if any) for why a package/version was removed.
+    /// </summary>
+    IReadOnlySet<PackageRemovalReason>? RemovalReasons { get; }
```
pmalmsten, line 32: This property is only relevant for the case of a package version having been removed. Given that, does it need to be exposed at the interface level - i.e., could users do a [pattern-matching type test](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#type-tests) to get access to this field only if/when the package version was removed?

src/Shared/Model/Enums/PackageRemovalReason.cs
```diff
@@ -0,0 +1,24 @@
...
+
+/// <summary>
+/// Enum for information about when a package no longer exists on a package manager.
+/// </summary>
+public enum PackageRemovalReason
+{
+    /// <summary>
+    /// When the version was unpublished from the repository/registry by someone.
+    /// </summary>
+    VersionUnpublished,
```
pmalmsten, line 13: Should we have different Enums when checking package existence vs. package version existence? I know there would be some duplication, but if a user asks OSSGadget about the existence of a package, that method would never return `VersionUnpublished`, correct? As this PR stands, the type system would not help the user avoid checking for an impossible value, as it could if there were separate enums for package vs. package version

jpinz: Sounds good! Addressed and resolving.

src/Shared/PackageManagers/BaseProjectManager.cs
jpinz: I just implemented it as an extension method in our project which for the time being is exactly what we need.

gfs:
I'll re-review today.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -364,14 +384,21 @@ public virtual async Task<bool> PackageVersionExistsAsync(PackageURL purl, bool
...
-            throw new NotImplementedException("BaseProjectManager does not implement PackageVersionPulled.");
+            bool exists = await PackageVersionExistsAsync(purl, useCache);
+
+            if (exists)
+            {
+                return new PackageVersionExists();
+            }
+
+            return new PackageVersionNotFound();
         }
```
pmalmsten, line 402: I was thinking a bit more about the default behavior you have here - in the event that a package version was unpublished, `PackageVersionExistsAsync` should return `false`, correct?

I ask because I want to double check the extensibility of this API contract if/when we add support for `PackageVersionRemoved` going forward. One question I'm thinking about is whether `PackageVersionRemoved` should be a subtype of `PackageVersionNotFound`, or whether they should be independent (as currently written in this PR). 

If `PackageVersionExistsAsync` does in fact return `false` for the unpublished/removed cases, then the default implementation of this method will return `PackageVersionNotFound` for now. When we come back and start adding support for `PackageVersionRemoved` to additional types of packages, it seems like `PackageVersionRemoved` would need to be a subtype of `PackageVersionNotFound`, to avoid breaking any code that relies on the default behavior here of returning `PackageVersionNotFound` in that case (i.e. if a caller has an expression like `result is PackageVersionNotFound`, and that expression evaluated to true for a given unpublished Cargo package version, it should continue to evaluate to true after we add support for `PackageVersionRemoved` for Cargo).

pmalmsten: Same discussion applies to `DetailedPackageExistsAsync` as well.

jpinz: This makes sense to me. Removed implies it wasn't found so I can make it a subset.

jpinz: Give it a look when you can, but I think I addressed what you were getting at, and added unit tests for it as well!

src/Shared/Model/PackageExistence/PackageVersionExistence.cs
```diff
@@ -0,0 +1,34 @@
...
+    public bool HasEverExisted => false;
+}
+
+/// <summary>
+/// Represents a package version that was removed, and why.
+/// </summary>
+/// <param name="RemovalReasons">The reasons (if any) for why a package was removed.</param>
+public record PackageVersionRemoved(IReadOnlySet<PackageVersionRemovalReason> RemovalReasons) : PackageVersionNotFound
+{
+    public new bool HasEverExisted => true;
```
gfs, line 33: I don't think you want to use new here.

The value of `HasEverExisted` will vary depending on the type of the variable, so if these are cast as `PackageVersionNotFound` objects they will actually end up with `HasEverExisted` false from the base `PackageVersionNotFound`. 

For Example:

image

src/Shared/Model/PackageExistence/PackageExistence.cs
```diff
@@ -0,0 +1,34 @@
...
+    public bool HasEverExisted => false;
+}
+
+/// <summary>
+/// Represents a package that was removed, and why.
+/// </summary>
+/// <param name="RemovalReasons">The reasons (if any) for why a package was removed.</param>
+public record PackageRemoved(IReadOnlySet<PackageRemovalReason> RemovalReasons) : PackageNotFound
+{
+    public new bool HasEverExisted => true;
```
gfs, line 33: Same issue with using new.

gfs (changes requested):
See comments about use of new. I suspect this is not the behavior you want.

src/Shared/Model/PackageExistence/PackageVersionExistence.cs
jpinz: Thanks for explaining! I didn't really have a good understanding of what new did, it's just what my IDE suggested I do.

jpinz: Done!

gfs: 👍. I think new is pretty confusing, I don't think I've yet found the case where it's actually what I want.

src/oss-tests/PackageExistenceTests.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        IPackageExistence packageExistenceRemoved = new PackageRemoved(new HashSet<PackageRemovalReason>(new[]
+        {
+            PackageRemovalReason.PackageUnpublished
+        }));
+
+        packageExistenceRemoved.Should().BeAssignableTo<PackageNotFound>()
+            .Subject.HasEverExisted.Should().BeTrue();
+
+        packageExistenceRemoved.HasEverExisted.Should().BeTrue();
```
pmalmsten, line 29: Small nit: this assertion appears redundant with the one on line 27. Looks like the same redundancy appeared in the below test case as well.

pmalmsten (approved):
The recent changes look great - thanks for the improvements! Left you one small nit, but otherwise ship it 👍

src/oss-tests/PackageExistenceTests.cs
jpinz: Yeah I guess it's kinda redundant but I added it for sanity sake. I don't particularly see a problem with having it, but I understand the confusion/redundancy reason to remove it.

gfs (approved):

PR #354, opened 2022-09-30
Remove HasEverExisted and WasRemoved flags from IPackageExistence

jpinz:
Remove HasEverExisted and WasRemoved flags from IPackageExistence as we can't be sure those values are correct as the registries aren't 100% immutable. We can't guarantee that HasEverExisted would be correct in those cases.

ex: this package [@softvisio/webview](https://registry.npmjs.org/@softvisio/webview) did exist, but now 404s.


pmalmsten (approved):

gfs (approved):

gfs:
I think I missed the failing tests in this PR. What's up with that?

PR #358, opened 2022-10-20
Nuget package versions are case insensitive

michelleqyun:
Package versions in NuGet are case insensitive, and NuGet's GET package versions API returns a list of all package versions in lowercase. OSSGadget's NuGetProjectManager should lowercase a package version before comparing it against the returned list of package versions.

### DIFF
--- a/src/Shared/PackageManagers/NuGetProjectManager.cs
+++ b/src/Shared/PackageManagers/NuGetProjectManager.cs
@@ -20,6 +20,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
     using System.Net.Http;
     using System.Text.Json;
     using System.Threading.Tasks;
+    using Microsoft.Extensions.Primitives;
 
     public class NuGetProjectManager : TypedManager<NuGetPackageVersionMetadata, NuGetProjectManager.NuGetArtifactType>
     {
@@ -149,6 +150,11 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                                            throw new InvalidOperationException($"Can't find the latest version of {purl}");
                     packageVersion = latestVersion;
                 }
+                // NuGet package versions are case insensitive
+                else
+                {
+                    packageVersion = packageVersion.ToLowerInvariant();
+                }
 
                 // Construct a new PackageURL that's guaranteed to have a version.
                 PackageURL purlWithVersion = new (purl.Type, purl.Namespace, packageName, packageVersion, purl.Qualifiers, purl.Subpath);
@@ -177,8 +183,8 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                                     throw new InvalidOperationException($"Can't find the latest version of {purl}");;
 
             // Construct a new PackageURL that's guaranteed to have a version, the latest version is used if no version was provided.
-            PackageURL purlWithVersion = !string.IsNullOrWhiteSpace(purl.Version) ? 
-                purl : new PackageURL(purl.Type, purl.Namespace, purl.Name, latestVersion, purl.Qualifiers, purl.Subpath);
+            string guaranteedVersion = !string.IsNullOrWhiteSpace(purl.Version) ? purl.Version.ToLowerInvariant() : latestVersion;
+            PackageURL purlWithVersion = new PackageURL(purl.Type, purl.Namespace, purl.Name, guaranteedVersion, purl.Qualifiers, purl.Subpath);
 
             NuGetPackageVersionMetadata? packageVersionMetadata =
                 await Actions.GetMetadataAsync(purlWithVersion, useCache: useCache);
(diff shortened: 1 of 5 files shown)


src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -207,7 +213,28 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
+
+            if(purl.Version.IsBlank())
+            {
+                Logger.Trace("Provided PackageURL version was null or blank.");
+                return false;
+            }
 
+            // NuGet packages are case insensitive. Convert purl to lowercase before comparing against package versions list. 
+            var versionLowercase = purl.Version.ToLowerInvariant();
+            return (await EnumerateVersionsAsync(purl, useCache)).Contains(versionLowercase);
```
gfs, line 235: Rather than converting the string to lowercase since the lowered string is only used in this call I think it would be better to use the Linq Contains with the appropriate string comparer. 

https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparer.invariantcultureignorecase?view=net-7.0

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -177,8 +183,8 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
                                     throw new InvalidOperationException($"Can't find the latest version of {purl}");;
 
             // Construct a new PackageURL that's guaranteed to have a version, the latest version is used if no version was provided.
-            PackageURL purlWithVersion = !string.IsNullOrWhiteSpace(purl.Version) ? 
-                purl : new PackageURL(purl.Type, purl.Namespace, purl.Name, latestVersion, purl.Qualifiers, purl.Subpath);
+            string guaranteedVersion = !string.IsNullOrWhiteSpace(purl.Version) ? purl.Version.ToLowerInvariant() : latestVersion;
```
gfs, line 186: I don't think this is necessary. The spec indicates that version checking is case insensitive. Not that it requires lowercase. 

Can you provide an example of when this change is required?

gfs (changes requested):
Thank you for the contribution. 

I'd prefer the first check be implemented with the string compare per my inline comment.

I'm not clear on the need for the second change that constructs the URL. 

The comparison of strings per the documentation is case insensitive so there's no need to modify the version itself as any nuget server should be doing a case insensitive comparison. 

> NuGetVersion uses case insensitive string comparisons for pre-release components. This means that 1.0.0-alpha and 1.0.0-Alpha are equal.

https://learn.microsoft.com/en-us/nuget/concepts/package-versioning

In order for me to accept and merge your contribution you will need to follow the bot's instructions to accept the CLA.

michelleqyun:
@microsoft-github-policy-service agree [company="Microsoft"]

michelleqyun:
@microsoft-github-policy-service agree company="Microsoft"

gfs:
@michelleqyun are you a Microsoft employee? if you associate your github account with the Microsoft organization you should be able to skip the CLA.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -221,9 +248,10 @@ private async Task UpdateVersionMetadata(PackageMetadata metadata, NuGetPackageV
             }
 
             string nameLowercase = packageVersionPackageVersionMetadata.Name.ToLowerInvariant();
+            string versionLowercase = metadata.PackageVersion.ToLowerInvariant();
```
gfs, line 251: This lowercasing is not needed for the test cases you added. Do you have an example of when it would be?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -221,9 +248,10 @@ private async Task UpdateVersionMetadata(PackageMetadata metadata, NuGetPackageV
...
 
             // Set the version specific URI values.
-            metadata.VersionUri = $"{metadata.PackageManagerUri}/packages/{nameLowercase}/{metadata.PackageVersion}";
+            metadata.VersionUri = $"{metadata.PackageManagerUri}/packages/{nameLowercase}/{versionLowercase}";
```
gfs, line 254: The metadata reporting URI ends up differing from the actual URI used for the query which is not lowercased.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -221,9 +248,10 @@ private async Task UpdateVersionMetadata(PackageMetadata metadata, NuGetPackageV
             }
 
             string nameLowercase = packageVersionPackageVersionMetadata.Name.ToLowerInvariant();
```
gfs, line 250: This doesn't appear to be a chance you introduced, but this also doesn't seem to be needed and causes the metadata to differ from the actual query used.

gfs (changes requested):
See comments. I had to make the changes to double check that they would work. I pushed them here: https://github.com/microsoft/OSSGadget/tree/pr/358-suggested, along with a new test for case insensitive package names, from which you can git cherry-pick.

src/Shared/PackageManagers/NuGetProjectManager.cs
michelleqyun: No, not particularly. I noticed the package version name was lowercased, so I just went ahead and did the same for the version.

gfs (approved):

PR #359, opened 2022-10-21
Add Verified to PackageMetadata, currently only used by NuGet.

jpinz:
It states if the specified package is verified as a part of the prefix/scope/namespace it is in.
https://learn.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation
Some packages in a NuGet prefix could not be a part of the prefix itself. So this adds the ability to assert that it is verified to be a part of the namespace.

For example: 
This package [`Microsoft.CST.OSSGadget.Shared`](https://www.nuget.org/packages/Microsoft.CST.OSSGadget.Shared/) is verified to be a part of the `Microsoft.` prefix (Look under the name for the checkmark)
But, the package [`Microsoft.Office.Core.Excel`](https://www.nuget.org/packages/Microsoft.Office.Core.Excel/) is not verified as a part of the `Microsoft.` prefix. Same for [`Microsoft.Azure.Functions.Library`](https://www.nuget.org/packages/Microsoft.Azure.Functions.Library/) and plenty of others.

Also, this PR fixes the build failures due to the NuGet.Protocol version bump.

### DIFF
--- a/src/Shared/Contracts/IManagerPackageActions.cs
+++ b/src/Shared/Contracts/IManagerPackageActions.cs
@@ -75,5 +75,5 @@ public interface IManagerPackageActions<T> where T : IManagerPackageVersionMetad
     /// <param name="useCache">If the cache should be checked if the package is in a reserved namespace.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to be used in the method call, defaults to <see cref="CancellationToken.None"/>.</param>
     /// <returns>True if the package is verified to be in a reserved namespace, false if not.</returns>
-    Task<bool> GetReservedNamespaceAsync(PackageURL packageUrl, bool useCache = true, CancellationToken cancellationToken = default);
+    Task<bool> GetHasReservedNamespaceAsync(PackageURL packageUrl, bool useCache = true, CancellationToken cancellationToken = default);
 } 
--- a/src/Shared/Contracts/IManagerPackageActions.cs
+++ b/src/Shared/Contracts/IManagerPackageActions.cs
@@ -67,4 +67,13 @@ public interface IManagerPackageActions<T> where T : IManagerPackageVersionMetad
     /// <returns>The <see cref="T"/> for this package version. Or null if none was found.</returns>
     /// <exception cref="ArgumentException">Thrown if there was no version in <paramref name="packageUrl"/>.</exception>
     Task<T?> GetMetadataAsync(PackageURL packageUrl, bool useCache = true, CancellationToken cancellationToken = default);
+    
+    /// <summary>
+    /// Gets if the package is in a reserved namespace.
+    /// </summary>
+    /// <param name="packageUrl">The <see cref="PackageURL"/>.</param>
+    /// <param name="useCache">If the cache should be checked if the package is in a reserved namespace.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to be used in the method call, defaults to <see cref="CancellationToken.None"/>.</param>
+    /// <returns>True if the package is verified to be in a reserved namespace, false if not.</returns>
+    Task<bool> GetReservedNamespaceAsync(PackageURL packageUrl, bool useCache = true, CancellationToken cancellationToken = default);
 } 
--- a/src/Shared/Model/Metadata/NuGetPackageVersionMetadata.cs
+++ b/src/Shared/Model/Metadata/NuGetPackageVersionMetadata.cs
@@ -91,10 +91,10 @@ public record NuGetPackageVersionMetadata : IManagerPackageVersionMetadata
     public bool IsListed { get; init; }
 
     /// <summary>
-    /// Gets if the package is verified within the prefix it is a part of.
+    /// Gets if the package is a part of a reserved prefix.
     /// </summary>
     [JsonProperty(PropertyName = JsonProperties.PrefixReserved)]
-    public bool Verified { get; init; }
+    public bool PrefixReserved { get; init; }
 
     [JsonIgnore]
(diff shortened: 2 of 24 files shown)


pmalmsten (approved):

scovetta:
Suggestion -- instead of "verified", maybe something like "isReservedNamespace" or something more specific? Also, if *only* NuGet has it (or exposes it), I'm not sure it should be part of the base model.

I guess `null` would mean "no information available"?

scovetta (dismissed):
See my other comment -- I think Verified is the wrong word here -- probably not PrefixReserved either, since that's an implementation detail -- maybe NamespaceReserved or something like that?

Also, if only NuGet implements/exposes this, should it be in the base model or just the NuGetPackageManager? Are there other fields that we have only in the child models?

src/Shared/Model/Metadata/NuGetPackageVersionMetadata.cs
```diff
@@ -90,8 +90,11 @@ public record NuGetPackageVersionMetadata : IManagerPackageVersionMetadata
     [JsonProperty(PropertyName = JsonProperties.Listed)]
     public bool IsListed { get; init; }
 
+    /// <summary>
+    /// Gets if the package is verified within the prefix it is a part of.
+    /// </summary>
     [JsonProperty(PropertyName = JsonProperties.PrefixReserved)]
-    public bool PrefixReserved { get; init; }
+    public bool Verified { get; init; }
```
gfs, line 97: I don't think this is a good name for this field. Verified on its own to me would indicate that we have done some verification of the package. Prefix reserved seems to be much more accurately describing the state.

src/Shared/Model/PackageMetadata.cs
```diff
@@ -135,6 +135,13 @@ public class PackageMetadata
         [JsonProperty(PropertyName = "active_flag", NullValueHandling = NullValueHandling.Ignore)]
         public bool? Active { get; set; }
 
+        /// <summary>
+        /// Gets or sets if the package is verified to be a part of the prefix/scope/namespace it is a part of.
+        /// And not grandfathered in, or a 3rd party package/part of a 3rd party namespace within the namespace. 
+        /// </summary>
+        [JsonProperty(PropertyName = "verified", NullValueHandling = NullValueHandling.Ignore)]
```
gfs, line 142: Json property name differs from NugetPackageVersionMetadata. Same concern that I don't believe this is the right terminology.

gfs (changes requested):

jpinz:
@gfs @scovetta I reverted the changes for the NuGet metadata related tweaks I made. And instead implemented a method on the NuGetProjectManager that just returns a boolean whether or not the package is verified to be a part of a reserved prefix.

gfs:
Can you add a test case or two for this new functionality?

jpinz:
> Can you add a test case or two for this new functionality?

@gfs done

src/Shared/Contracts/IManagerPackageActions.cs
```diff
@@ -67,4 +67,13 @@ public interface IManagerPackageActions<T> where T : IManagerPackageVersionMetad
...
     Task<T?> GetMetadataAsync(PackageURL packageUrl, bool useCache = true, CancellationToken cancellationToken = default);
+    
+    /// <summary>
+    /// Gets if the package is in a reserved namespace.
+    /// </summary>
+    /// <param name="packageUrl">The <see cref="PackageURL"/>.</param>
+    /// <param name="useCache">If the cache should be checked if the package is in a reserved namespace.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to be used in the method call, defaults to <see cref="CancellationToken.None"/>.</param>
+    /// <returns>True if the package is verified to be in a reserved namespace, false if not.</returns>
+    Task<bool> GetReservedNamespaceAsync(PackageURL packageUrl, bool useCache = true, CancellationToken cancellationToken = default);
```
gfs, line 78: Can we rename this to something like "GetHasReservedNamespaceAsync"? This returns bool but based on the method name I'd assume it returns the namespace itself.

src/Shared/PackageActions/NuGetPackageActions.cs
```diff
@@ -167,4 +167,24 @@ public class NuGetPackageActions : IManagerPackageActions<NuGetPackageVersionMet
 
         return packageVersion != null ? new NuGetPackageVersionMetadata(packageVersion) : null;
     }
+
+    /// <inheritdoc />
+    public async Task<bool> GetReservedNamespaceAsync(PackageURL packageUrl, bool useCache = true, CancellationToken cancellationToken = default)
```
gfs, line 172: Same comment as above with name of method vs return type

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -169,6 +169,17 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
             return new Uri($"{ENV_NUGET_HOMEPAGE}/{purl?.Name}");
         }
+
+        /// <summary>
+        /// Gets if the package is a part of a reserved prefix.
+        /// </summary>
+        /// <param name="purl">The package url to check.</param>
+        /// <param name="useCache">If the cache should be used.</param>
+        /// <returns>True if the package is verified to be in a reserved prefix, false if not.</returns>
+        public async Task<bool> GetPackagePrefixReservedAsync(PackageURL purl, bool useCache = true)
```
gfs, line 179: same comment where the method name makes me expect a string or some meaningful value but this returns a bool.

gfs (commented):
Minor nits on the names for things added to the interface.

gfs (approved):

PR #361, opened 2022-10-26
Improve PyPI GetArtifactDownloadUris.

jpinz:
- Adds optional property UploadTime to ArtifactUri<T>.
- Implement PyPI GetArtifactDownloadUris entirely, it now gets all associated artifacts. 
- PyPI uploadTime is now populated by the upload_time value of the file with the oldest upload_time to get the oldest uploadTime (publish time) for this package version. 
- Implement tests for these changes.


src/Shared/PackageManagers/PyPIProjectManager.cs
```diff
@@ -42,13 +43,50 @@ public class PyPIProjectManager : TypedManager<IManagerPackageVersionMetadata, P
         /// <inheritdoc />
         public override IEnumerable<ArtifactUri<PyPIArtifactType>> GetArtifactDownloadUris(PackageURL purl)
         {
-            string feedUrl = (purl.Qualifiers?["repository_url"] ?? ENV_PYPI_ENDPOINT).EnsureTrailingSlash();
+            Check.NotNull(nameof(purl.Version), purl.Version);
+            // TODO: Is this OK to do?
+            return this.GetArtifactDownloadUrisAsync(purl).ToEnumerable();
```
jpinz, line 48: @gfs I've never done anything like this before, unsure if there's a better solution.
One thought I had was just making GetArtifactDownloadUris async to begin with, but didn't want to do any contract breaking. 
Let me know what you think.

gfs: I don't think I've used that APi before, but from the description - and since the tests seem to be functional, I think its fine.

gfs, line 48: Nit: Style preference is to not use `this` unless it is required to deconflict.

src/oss-tests/ProjectManagerTests/PyPIProjectManagerTests.cs
```diff
@@ -161,5 +214,21 @@ public async Task GetArtifactDownloadUrisSucceeds_Async(string purlString, strin
...
+
+        /// <summary>
+        /// Helper method to get the <see cref="PyPIProjectManager.PyPIArtifactType"/> from a <see cref="Uri"/>.
+        /// </summary>
+        /// <param name="uri">The <see cref="Uri"/> to use.</param>
+        /// <returns>The <see cref="PyPIProjectManager.PyPIArtifactType"/> from the <see cref="Uri"/>.</returns>
+        private static PyPIProjectManager.PyPIArtifactType ExtensionToType(Uri uri)
+        {
+            string extension = uri.GetExtension();
+            return extension switch
```
gfs, line 226: Nit: Could be a lambda expression

```csharp
private static PyPIProjectManager.PyPIArtifactType ExtensionToType(Uri uri) => uri.GetExtension() switch
{
    ".tar.gz" => PyPIProjectManager.PyPIArtifactType.Tarball,
    ".whl" => PyPIProjectManager.PyPIArtifactType.Wheel,
    _ => PyPIProjectManager.PyPIArtifactType.Unknown
};
```

src/Shared/PackageManagers/PyPIProjectManager.cs
```diff
@@ -292,54 +344,54 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
             // if we found any version at all, get the information.
             if (metadata.PackageVersion is not null)
             {
-                Version versionToGet = new(metadata.PackageVersion);
-                JsonElement? versionElement = GetVersionElement(contentJSON, versionToGet);
-                if (versionElement is not null)
+                JsonElement.ArrayEnumerator?
+                    urlsArray = OssUtilities.GetJSONEnumerator(root.GetProperty("urls"));
```
gfs, line 348: Formatting, combine with previous line.

gfs (commented):
Overall looks good. See comments for formatting.

gfs (approved):
The .Result pattern is more familiar to me. 

Good idea to obsolete the sync version.

PR #362, opened 2022-10-27
Add NuGet badge to README.md

jpinz:


README.md
```diff
@@ -1,3 +1,4 @@
+![Nuget](https://img.shields.io/nuget/v/Microsoft.CST.OSSGadget.Shared)
```
gfs, line 1:
```suggestion
[![Nuget](https://img.shields.io/nuget/v/Microsoft.CST.OSSGadget.Shared)](https://www.nuget.org/packages/Microsoft.CST.OSSGadget.Shared)
```
I think you want to do something like this. If you don't have an extra link markdown around the badge it is clickable but it will just open the badge icon in the browser.

gfs (approved):
Changes are fine just to cause a build snap. But the badge could be improved per comment.

PR #363, opened 2022-10-27
Add GetPackagesFromOwnerAsync for the TypedManager.

jpinz:
Implement for Npm and PyPI. Can't figure out NuGet easily so it just throws a NotImplementedException.


gfs (approved):
I think this is fine. I could take a look at the nuget implementation tomorrow if you want to wait for that.

jpinz:
> I think this is fine. I could take a look at the nuget implementation tomorrow if you want to wait for that.

We primarily needed this for npm, so no rush for NuGet. I think this is ok for now.

PR #367, opened 2022-11-15
Change Environment Variables to be non-static + public

gfs:
Update Environment helper to match only modifying public statics


jpinz (approved):
LGTM :shipit:

PR #375, opened 2022-12-02
Update Cargo, Maven, and Golang Project Managers to extend TypedManager class

brbayes-msft:
This pull request updates the Cargo, Maven, and Golang project managers to extend the TypedManager base class.  This introduces functionality for exposing the download artifact uris through a method.  This change also updates the Maven project manager to appropriately support custom feed urls for downloads.


jpinz (approved):
Looks good to me!

PR #376, opened 2022-12-05
Add missing SDK for NuGet builds

brbayes-msft:
The build is currently failing to build NuGet packages on the code signing step due to a missing .NET SDK.  This change adds in that missing SDK prior to the code signing.


jpinz (approved):
LGTM :shipit:

PR #386, opened 2023-01-09
Add basic support for Conda

gpcastro:
We only need check for package existence at this point, but I'm also adding the package download code as part of this change.

### DIFF
--- a/src/Shared/PackageManagers/CondaProjectManager.cs
+++ b/src/Shared/PackageManagers/CondaProjectManager.cs
@@ -1,0 +1,138 @@
+﻿// Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+
+namespace Microsoft.CST.OpenSource.PackageManagers
+{
+    using Contracts;
+    using Microsoft.CST.OpenSource.Helpers;
+    using Microsoft.CST.OpenSource.Model;
+    using Microsoft.CST.OpenSource.PackageActions;
+    using PackageUrl;
+    using System;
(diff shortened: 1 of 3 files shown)


src/Shared/PackageManagers/CondaProjectManager.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// <inheritdoc />
+        public override async IAsyncEnumerable<ArtifactUri<CondaArtifactType>> GetArtifactDownloadUrisAsync(PackageURL purl, bool useCache = true)
+        {
+            Check.NotNull(nameof(purl), purl);
+            if (!IsValidPackageUrlForConda(purl))
+            {
+                throw new ArgumentException("Invalid package URL for Conda.");
+            }
+
+            string? type = purl?.Qualifiers?.GetValueOrDefault(TypeQualifier);
```
brbayes-msft, line 50: I know that the file type is a part of the package url spec for Conda.  I am wondering about perhaps returning all relevant download urls, and then allowing the caller to filter to which file type they want.  Is it generally expected that only one of the conda or the tar file is generally available, or are both of them usually present?

gpcastro: From my analysis, tar.bz2 is more prevalent, in particular in the `free` channel. When the type is not provided, both values are returned (line 54 and 56 below)

brbayes-msft (approved):

PR #391, opened 2023-02-13
Fix some issues with the squat mutators

jpinz:
- Add a few more prefixes and suffixes for npm.
- Add substitution and suffix mutators for PyPi.
- Update the close letters mutator to add a close letter to the end of a string. We missed the node package sqliter from sqlite as a typosquat. 
- Fix substitution mutator.
- Fix Swap Order Of Letters mutator.
- Update FindSquatsTests.


brbayes-msft (approved):

gfs (approved):
LGTM.

PR #394, opened 2023-02-15
Improve the typosquat mutators

jpinz:
This PR aims to decrease the gap between original package name => mutated package name => mutations on the mutation including the original.
I added a test to show that it works. With some caveats like BitFlip mutator, and the separator and namespace related mutators not being able to easily get back to the original.


src/oss-find-squats-lib/MutateExtension.cs
```diff
@@ -43,9 +45,12 @@ public static class MutateExtension
         /// <summary>
         /// Common variations known uniquely for Nuget/C#. Excludes '.' because Nuget will match "Microsoft.CST.OAT." against "Microsoft.CST.OAT", giving false positives for the same package.
         /// </summary>
-        internal static IEnumerable<IMutator> NugetMutators { get; } = BaseMutators.Where(x => x is not UnicodeHomoglyphMutator and not SuffixMutator)
+        internal static IEnumerable<IMutator> NugetMutators { get; } = BaseMutators.Where(x => 
+                x is not UnicodeHomoglyphMutator and not SuffixMutator and not CloseLettersMutator and not DoubleHitMutator)
```
gfs, line 49: Why remove the double hit mutator from nuget?

src/oss-find-squats-lib/Mutators/RemoveSeparatedSectionMutator.cs
```diff
@@ -9,13 +9,13 @@ namespace Microsoft.CST.OpenSource.FindSquats.Mutators
     /// <summary>
     /// Generates mutations for if a character was added after a separator was used.
```
gfs, line 10: Is this comment still correct? It seems the purpose of this mutator was changed.

src/oss-find-squats-lib/Mutators/SeparatorMutator.cs
```diff
@@ -12,7 +12,10 @@ namespace Microsoft.CST.OpenSource.FindSquats.Mutators
     /// </summary>
     public class SeparatorMutator : IMutator
     {
-        public MutatorType Kind { get; } = MutatorType.Separator;
+        /// <summary>
+        /// Gets the MutatorType as <see cref="MutatorType.SeparatorRemoved"/>. It can however also be <see cref="MutatorType.SeparatorChanged"/>.
```
gfs, line 16: What does this comment mean?

gfs (commented):
Generally looks good. I noticed perhaps a couple oversights in the changes here.

src/oss-find-squats-lib/MutateExtension.cs
jpinz: We remove it as we are customizing it from the `BaseMutators` implementation. Similarly to how we were doing it for the `SuffixMutator`

src/oss-find-squats-lib/Mutators/RemoveSeparatedSectionMutator.cs
jpinz: Thanks for the catch! Fixed.

src/oss-find-squats-lib/Mutators/SeparatorMutator.cs
jpinz: So previously we considered both Removing a separator, and Changing a separator as a Separator Mutator. As `Kind` only accepts one for the class, I had picked SeparatorRemoved, but this doesn't make a lot of sense, so I will go split it into two mutators.

gfs (approved):
I have one minor recommendation around using the chained replace + insert. That will result in an additional string allocation for the remove before the insert that is just a throwaway value. It's possible to do without the interim allocations using string interpolation and slices per my comment.

src/Shared/Helpers/StringExtensions.cs
```diff
@@ -61,5 +61,29 @@ public static string EnsureTrailingSlash(this string url)
...
+        /// <summary>
+        /// Returns the input string <paramref name="source"/> with the first instance of <paramref name="oldValue"/>
+        /// being replaced with <paramref name="newValue"/>.
+        /// </summary>
+        /// <remarks>If the string doesn't have an instance of <paramref name="oldValue"/>, no changes are made.</remarks>
+        public static string ReplaceAtStart(this string source, string oldValue, string newValue)
+        {
+            int place = source.IndexOf(oldValue);
+    
+            return place == -1 ? source : source.Remove(place, oldValue.Length).Insert(place, newValue);
```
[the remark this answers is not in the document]
jpinz: Done!

PR #395, opened 2023-02-16
Checking for NPM or PyPi package version existence checks only one version

michelleqyun:
The previous implementation of listing all package versions and checking if the current version is included in that list is inefficient and uses significant memory in runtime. This PR changes the NPM, NuGet, and PyPi package managers to check the registry API endpoints only for the specific package version.


src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -148,6 +148,28 @@ public override async Task<bool> PackageExistsAsync(PackageURL purl, bool useCac
...
+                return false;
+            }
+
+            if (purl.Version.IsBlank())
+            {
+                Logger.Trace("Provided PackageURL version was null or blank.");
+                return false;
+            }
+
+            string packageName = purl.HasNamespace() ? $"{purl.GetNamespaceFormatted()}/{purl.Name}/{purl.Version}" : $"{purl.Name}/{purl.Version}";
```
jpinz, line 167: nit: not technically the `packageName` I would probably split this up to just the `packageName` declaration and finish the rest of the path in the actual query itself.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -420,7 +420,6 @@ public virtual async Task<IPackageExistence> DetailedPackageVersionExistsAsync(P
         /// </summary>
         /// <param name="purl">The <see cref="PackageURL"/> to get the metadata for.</param>
         /// <param name="useCache">If the metadata should be retrieved from the cache, if it is available.</param>
-        /// <remarks>If no version specified, defaults to latest version.</remarks>
```
jpinz, line 423: Why remove this? Pretty sure this is still the case.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -552,10 +574,14 @@ public override async Task<IPackageExistence> DetailedPackageVersionExistsAsync(
                     "Cannot call DetailedPackageVersionExists on a purl without a version. Call DetailedPackageExists.");
             }
 
-            string? content = await GetMetadataAsync(purl, useCache);
-            if (string.IsNullOrEmpty(content)) { return new PackageVersionNotFound(); }
+            bool packageVersionCurrentlyExists = await PackageVersionExistsAsync(purl, useCache);
+            if (packageVersionCurrentlyExists) { return new PackageVersionExists(); }
+            
+            // if version isn't currently listed, check for other kinds of existence
+            string? packageContent = await GetMetadataAsync(purl, useCache);
```
jpinz, line 581: nit: any reason you changed `content` to `packageContent`?

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -591,7 +617,7 @@ public override async Task<IPackageExistence> DetailedPackageVersionExistsAsync(
             {
                 return new PackageVersionRemoved(removalReasons);
             }
-
+
```
jpinz, line 620: nit: remove

src/Shared/PackageManagers/PyPIProjectManager.cs
```diff
@@ -190,6 +190,28 @@ public override async Task<bool> PackageExistsAsync(PackageURL purl, bool useCac
...
+                return false;
+            }
+
+            if (purl.Version.IsBlank())
+            {
+                Logger.Trace("Provided PackageURL version was null or blank.");
+                return false;
+            }
+
+            string packageName = $"{purl.Name}/{purl.Version}";
```
jpinz, line 209: nit: not `packageName`, I would just do this in line for the http request.

src/oss-tests/ProjectManagerTests/NPMProjectManagerTests.cs
```diff
@@ -155,13 +156,15 @@ public async Task EnumerateVersionsSucceeds(string purlString, int count, string
         [DataRow("pkg:npm/example@0.0.0")]
         [DataRow("pkg:npm/rly-cli@0.0.2")]
         [DataRow("pkg:npm/lodash.js@0.0.1-security")]
+        [DataRow("pkg:npm/tslib@2.4.1")]
         public async Task PackageVersionExistsAsyncSucceeds(string purlString)
         {
+            NPMProjectManager projectManager = new(".");
             PackageURL purl = new(purlString);
 
-            Assert.IsTrue(await _projectManager.Object.PackageVersionExistsAsync(purl, useCache: false));
```
jpinz, line 162: Why did you change this?

src/oss-tests/ProjectManagerTests/PyPIProjectManagerTests.cs
```diff
@@ -147,7 +147,20 @@ public async Task EnumerateVersionsSucceeds(string purlString, int count, string
...
+        [DataRow("pkg:pypi/requests@2.27.1")]
+        public async Task PackageVersionExistsAsyncSucceeds(string purlString)
+        {
+            PyPIProjectManager projectManager = new(".");
+            PackageURL purl = new(purlString);
+
+            Assert.IsTrue(await projectManager.PackageVersionExistsAsync(purl, useCache: false));
+        }
+
+
```
jpinz, line 163: nit: extra line here.

src/oss-tests/ProjectManagerTests/PyPIProjectManagerTests.cs
```diff
@@ -147,7 +147,20 @@ public async Task EnumerateVersionsSucceeds(string purlString, int count, string
...
         }
-        
+
+        [DataTestMethod]
+        [DataRow("pkg:pypi/pandas@1.4.2")]
+        [DataRow("pkg:pypi/plotly@3.7.1")]
+        [DataRow("pkg:pypi/requests@2.27.1")]
+        public async Task PackageVersionExistsAsyncSucceeds(string purlString)
+        {
+            PyPIProjectManager projectManager = new(".");
```
jpinz, line 157: You shouldn't have to create a new projectManager. Should be able to use `_projectManager`

src/oss-tests/ProjectManagerTests/PyPIProjectManagerTests.cs
```diff
@@ -163,7 +176,7 @@ public async Task DetailedPackageExistsAsync_WorksAsExpected(string purlString,
         [DataTestMethod]
         [DataRow("pkg:pypi/pandas@1.4.2", true)]
         [DataRow("pkg:pypi/pandas@12.34.56.78", false)]
-        [DataRow("pkg:pypi/plotly@3.7.1", true)]
+        [DataRow("pkg:pypi/plotly@5.7.0", true)]
```
jpinz, line 179: why did you change the version?

jpinz (changes requested):

src/Shared/PackageManagers/BaseProjectManager.cs
michelleqyun: The default implementation in BaseProjectManager.DetailedPackageVersionExistsAsync() is a call to PackageVersionExistsAsync(), which will exist if there is no version provided.

src/Shared/PackageManagers/NPMProjectManager.cs
gfs: Git doesn't track actual changes it tracks differences. If you delete something and add exactly the same thing git doesn't detect a difference. This seems to show an empty line deleted and a line with white space but no content added.

src/oss-tests/ProjectManagerTests/PyPIProjectManagerTests.cs
michelleqyun: The http responses in these unit tests rely on a mock HttpMessageHandler, which retrieves pre-saved json files in the /TestData/PyPI folder (rather than actually fetching from server). I changed the version here to use an existing json file rather than adding another to the test data folder.

michelleqyun: The reason the unit test was passing before is because DetailedPackageExistsAsync() by default calls PackageExistsAsync(), which before this PR used to call the non-version specific endpoint. This would pull the pre-saved json test data for plotly, but once I changed PackageExistsAsync() to poll the specific version, it would fail because we don't have a pre-saved json file for plotly@3.7.1. Since we do already have test data for plotly@5.7.0, I changed it to that package version.

src/oss-tests/ProjectManagerTests/PyPIProjectManagerTests.cs
michelleqyun: Sure, I can also use _projectManager as long as I set up the http responses with the mock handler at the top of file. I wasn't sure if we wanted to do that because I noticed NPMProjectManager.PackageVersionExistsAsyncSucceeds() creates a new project manager. Do we want to follow the same pattern or are we good to diverge?

bhuwan-agarwal:
@michelleqyun , is this PR ready to be published?

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -148,6 +148,28 @@ public override async Task<bool> PackageExistsAsync(PackageURL purl, bool useCac
             return await CheckJsonCacheForPackage(httpClient, $"{ENV_NPM_API_ENDPOINT}/{packageName}", useCache);
         }
 
+        /// <inheritdoc />
+        public override async Task<bool> PackageVersionExistsAsync(PackageURL purl, bool useCache = true) // MICHELLE
```
bhuwan-agarwal, line 152: Will there be a follow up PR in Terrapin to use this method (with useCache = false) instead of `DetailedPackageVersionExistsAsync`?

michelleqyun: `DetailedPackageVersionExistsAsync` by default calls `PackageVersionExistsAsync` in the base implementation. Project managers that override with their own version of `DetailedPackageVersionExistsAsync` also include `PackageVersionExistsAsync` in their implementation.

bhuwan-agarwal: Does this mean by default all supported package managers will now try fetching the singular version before all the versions?

michelleqyun: No, only for package managers that override the `PackageVersionExistsAsync` method, which will be PyPI, NPM, and NuGet after this PR goes in. The default behavior of `PackageVersionExistsAsync` as defined in BaseProjectManager is to list all the versions.

bhuwan-agarwal:
@michelleqyun, what all is left in this PR?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -226,26 +226,48 @@ public async Task<bool> GetHasReservedNamespaceAsync(PackageURL purl, bool useCa
...
             if (purl is null)
+            {
+                Logger.Trace("Provided PackageURL was null.");
+                throw new ArgumentNullException(nameof(purl), "Provided PackageURL was null.");
+            }
+
+            HttpClient httpClient = CreateHttpClient();
+            
+            // NuGet packages are case insensitive
+            string endpoint = $"{RegistrationEndpoint}{purl.Name.ToLowerInvariant()}/index.json";
```
jpinz, line 242: if they are case insensitive, is there a reason why you lowercase the name anyways? Seems like we shouldn't have to.

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -226,26 +226,48 @@ public async Task<bool> GetHasReservedNamespaceAsync(PackageURL purl, bool useCa
...
                 Logger.Trace("Provided PackageURL version was null or blank.");
                 return false;
             }
 
-            // NuGet packages are case insensitive.
-            return (await EnumerateVersionsAsync(purl, useCache)).Contains(purl.Version, StringComparer.InvariantCultureIgnoreCase);
+            HttpClient httpClient = CreateHttpClient();
+
+            // NuGet packages are case insensitive
+            string endpoint = $"{NUGET_DEFAULT_CONTENT_ENDPOINT.EnsureTrailingSlash()}{purl.Name.ToLowerInvariant()}/{purl.Version.ToLowerInvariant()}/{purl.Name.ToLowerInvariant()}.nuspec";
```
jpinz, line 266: Same here about case insensitivity.

jpinz (approved):
Looks good to me! Just left 2 comments cause I was curious.

src/Shared/PackageManagers/NuGetProjectManager.cs
michelleqyun: The endpoints are all normalized to lowercase to reflect the fact that package names are case insensitive. E.g. https://api.nuget.org/v3-flatcontainer/**oWin**/index.json will return does not exist, when actually it should point to the same place as https://api.nuget.org/v3-flatcontainer/owin/index.json if NuGet package names are true to case insensitivity. So, we lowercase.

src/Shared/PackageManagers/NuGetProjectManager.cs
michelleqyun: Same as other comment.

PR #399, opened 2023-02-20
Improve oss-find-source for PyPI packages.

### LINKED ISSUE #398: Can not find source for tensorflow and torch package
I run `oss-find-source` in the built docker image. `tensorflow` and `torch` packages both have corresponding source code repositories as presented in their PyPI project page. But `oss-find-source` can not find their repositories:
```shell
$ docker run -it --rm ossgadget /bin/bash
root@b1c9bb6f7ea5:~# oss-find-source pkg:pypi/tensorflow pkg:pypi/torch

   ____   _____ _____    _____           _            _
  / __ \ / ____/ ____|  / ____|         | |          | |
 | |  | | (___| (___   | |  __  __ _  __| | __ _  ___| |_
 | |  | |\___ \\___ \  | | |_ |/ _` |/ _` |/ _` |/ _ \ __|
 | |__| |____) |___) | | |__| | (_| | (_| | (_| |  __/ |_
  \____/|_____/_____/   \_____|\__,_|\__,_|\__, |\___|\__|
                                            __/ |
                                           |___/
OSS Gadget - oss-find-source 0.0.0+ebd2e46ccf - github.com/Microsoft/OSSGadget
INFO  - No repositories were found after searching metadata.
WARN  - No repositories found for package pkg:pypi/tensorflow
INFO  - No repositories were found after searching metadata.
WARN  - No repositories found for package pkg:pypi/torch
```
After inspecting the code, I find that currently `oss-find-source` only searches properties with certain keys in the `project_urls` filed. But it is possible that other properties (e.g., the `Download` property in tensorflow package's metadata) also contain source code repository information.: 
https://github.com/microsoft/OSSGadget/blob/ebd2e46ccf880491910dec33530dfe067df9e0cb/src/Shared/PackageManagers/PyPIProjectManager.cs#L470
However, the source code repository information are located in the `Download` property in `tensorflow` and `torch` packages' metadata. You may try :
```shell
$ curl -s https://pypi.org/pypi/tensorflow/json | jq -r '.info.project_urls'
{
  "Download": "https://github.com/tensorflow/tensorflow/tags",
  "Homepage": "https://www.tensorflow.org/"
}
```

gaokai320:
Fix #398 
Currently, oss-find-source only searches properties with certain keys in the `project_urls` filed. But it is possible that other properties (e.g., the `Download` property in tensorflow package's metadata) also contain source code repository information.

### DIFF
--- a/src/Shared/PackageManagers/PyPIProjectManager.cs
+++ b/src/Shared/PackageManagers/PyPIProjectManager.cs
@@ -41,7 +41,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
             : base(actions ?? new NoOpPackageActions(), httpClientFactory ?? new DefaultHttpClientFactory(), directory)
         {
         }
-        
+
         /// <inheritdoc />
         public override async IAsyncEnumerable<ArtifactUri<PyPIArtifactType>> GetArtifactDownloadUrisAsync(PackageURL purl, bool useCache = true)
         {
@@ -88,7 +88,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
             {
                 throw new InvalidOperationException();
             }
-            
+
             HtmlParser parser = new();
             AngleSharp.Html.Dom.IHtmlDocument document = await parser.ParseDocumentAsync(html);
             foreach (AngleSharp.Dom.IElement packageSnippet in document.QuerySelectorAll("a.package-snippet"))
@@ -285,7 +285,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
             JsonElement root = contentJSON.RootElement;
 
             JsonElement infoElement = root.GetProperty("info");
-            
+
             metadata.LatestPackageVersion = OssUtilities.GetJSONPropertyStringIfExists(infoElement, "version"); // Ran in the root, always points to latest version.
 
             if (purl.Version.IsBlank() && metadata.LatestPackageVersion.IsNotBlank())
@@ -392,7 +392,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                             metadata.VersionUri = $"{ENV_PYPI_ENDPOINT}/project/{purl.Name}/{purl.Version}";
                             metadata.VersionDownloadUri = OssUtilities.GetJSONPropertyStringIfExists(url, "url");
                         }
-                        
+
                         string? uploadTime = OssUtilities.GetJSONPropertyStringIfExists(url, "upload_time");
                         if (uploadTime != null)
                         {
@@ -467,7 +467,6 @@ namespace Microsoft.CST.OpenSource.PackageManagers
             JsonDocument contentJSON = JsonDocument.Parse(metadata);
 
             List<string> possibleProperties = new() { "homepage", "home_page" };
-            List<string> searchPropertiesProjectURLs = new() { "Source", "Source Code", "homepage", "Bug Tracker", "Repository", "Bug Reports" };
 
             JsonElement infoJSON;
             try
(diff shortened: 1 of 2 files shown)


gaokai320:
@microsoft-github-policy-service agree [company="{your company}"]

gaokai320:
@microsoft-github-policy-service agree

gfs:
Thanks for the contribution. I'll review this later today.

src/Shared/PackageManagers/PyPIProjectManager.cs
```diff
@@ -467,7 +467,6 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
             JsonDocument contentJSON = JsonDocument.Parse(metadata);
 
             List<string> possibleProperties = new() { "homepage", "home_page" };
-            List<string> searchPropertiesProjectURLs = new() { "Source", "Source Code", "homepage", "Bug Tracker", "Repository", "Bug Reports" };
```
gfs, line 470: Per your report, for the repositories this isn't working for, the git url is held in the "download" property. It seems like the simpler solution would be to add "download" to the list of properties we search for. This seems like it might have some unintended side effects if there happen to be unrelated git repositories under a property we aren't anticipating.  Is there a reason why that wouldn't work?

src/Shared/PackageManagers/PyPIProjectManager.cs
```diff
@@ -496,17 +495,16 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
                     }
                     else if (property.Name.Equals("project_urls"))
                     {
-                        foreach (var propertyName in searchPropertiesProjectURLs)
+                       foreach (JsonProperty project_url in property.Value.EnumerateObject())
                         {
-                            if (property.Value.TryGetProperty(propertyName, out JsonElement projectUrl))
+                            string projectUrl = project_url.Value.ToString() ?? string.Empty;
```
gfs, line 500: Assuming we go this path, there's no need to call ExtractGitHubPackageURLs on string.empty, so this could be an if instead.

gfs:
@jpinz what do you think about this proposed change?

src/Shared/PackageManagers/PyPIProjectManager.cs
gaokai320: According to the reference of [`setup`](https://setuptools.pypa.io/en/latest/references/keywords.html), the `project_urls` filed is
> An arbitrary map of URL names to hyperlinks, allowing more extensible documentation of where various resources can be found than the simple url and download_url options provide.

So I think `project_urls` may contain code repository information in properties with arbitrary keys, such as Code, Issue Tracker, Documentation, etc. An example is `aws-satella` package. A conservative way is to maintain a list containing very possible keys. An aggressive way is to search all properties in the `project_urls` field.

src/Shared/PackageManagers/PyPIProjectManager.cs
gaokai320: I agree with you. Since I am not familiar with C#, I follow the code in Line 487. Shall I change both lines?

jpinz (approved):
Assuming it works as expected it looks good to me! Good catch with the project url stuff, much appreciated!

gfs (approved):

gaokai320:
https://github.com/microsoft/OSSGadget/blob/a62ff644f29a1e6bb40b53a729ba15e9eb141853/src/Shared/PackageManagers/PyPIProjectManager.cs#L470
I notice that currently the properties with keys `url`, `download_url`, or `project_urls` are searched. I am sorry for the confusion of the provided link: https://setuptools.pypa.io/en/latest/references/keywords.html. 
`url`, `download_url`, and `project_urls` are parameters in the `setup` function. They are used to generate distribution's metadata (i.e., the `info` property in the result of querying [PyPI's JSON API](https://warehouse.pypa.io/api-reference/json.html#get--pypi--project_name--json)). However, the `url` parameter in the `setup` function turns to the `home_page` property. You may refer the the [example](https://warehouse.pypa.io/api-reference/json.html#get--pypi--project_name--json) to see the properties that the returned JSON data contains.
So I think this line of code should be:
```c#
List<string> possibleLocationProperties = new() { "home_page", "download_url" }; 
```

gfs:
I see, thanks for the extra clarification. I can address shortly.

PR #400, opened 2023-02-21
Update squat logic to address bugs.

jpinz:
- Remove `SeparatorRemovedMutator` from the list of `BaseMutators` as it was causing problems and usually doesn't catch squats. Squats that would be caught by it are also caught by the prefix or suffix mutator.
- Overrode the `SeparatorChangedMutator` for PyPi as PyPi considers `-` and `_` to be the same character. So just removed `_` from the list. There might be a better way of doing this, but I haven't thought of one yet.
- Fixed issues with the `DuplicatorMutator` that was giving the original name back as a mutated one.


gfs (approved):
Looks simple enough.

PR #404, opened 2023-03-06
Fix some bugs

jpinz:
- Readd `RemoveSeparatedSectionMutator` to the list of defaults, but remove it specifically for NPM.
- Fix issues where mutations would be just a separator. EX: `pkg:npm/q` would make `pkg:npm/.` or `pkg:npm/typescript` could make `pkg:npm/-`
- Fix bug with scoped npm packages with the prefix mutator would add before the `@` resulting in mutations like `pkg:npm/-%40types/node` when that isn't possible. It now adds after the `@` but before the namespace begins, resulting in `pkg:npm/%40-types/node`

### DIFF
--- a/src/oss-find-squats-lib/Mutators/IMutator.cs
+++ b/src/oss-find-squats-lib/Mutators/IMutator.cs
@@ -42,12 +42,12 @@ namespace Microsoft.CST.OpenSource.FindSquats.Mutators
                     continue;
                 }
                 
-                if (mutation.Mutated.Length == 1 &&
-                    SeparatorRemovedMutator.DefaultSeparators.Contains(Convert.ToChar(mutation.Mutated)))
+                if (mutation.Mutated.Length == 1 && !char.IsLetterOrDigit(Convert.ToChar(mutation.Mutated)))
                 {
                     // Don't make mutations that are just one separator. i.e pkg:npm/. isn't valid.
                     continue;
                 }
+
                 if (hasNamespace)
                 {
                     yield return new Mutation(
--- a/src/oss-find-squats-lib/Mutators/IMutator.cs
+++ b/src/oss-find-squats-lib/Mutators/IMutator.cs
@@ -4,6 +4,7 @@ namespace Microsoft.CST.OpenSource.FindSquats.Mutators
 {
     using Extensions;
     using PackageUrl;
+    using System;
     using System.Collections.Generic;
     using System.Web;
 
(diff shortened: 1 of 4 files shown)


brbayes-msft (approved):

src/oss-find-squats-lib/Mutators/IMutator.cs
```diff
@@ -35,6 +36,12 @@ public IEnumerable<Mutation> Generate(PackageURL arg)
             bool hasNamespace = arg.HasNamespace();
             foreach (Mutation mutation in Generate(hasNamespace ? arg.Namespace : arg.Name))
             {
+                if (mutation.Mutated.Length == 1 &&
+                    SeparatorRemovedMutator.DefaultSeparators.Contains(Convert.ToChar(mutation.Mutated)))
```
gfs, line 40: I wonder if checking if the single character is alphanumeric might be better here rather than if it's just one of the default separators. Do we know of any manager where a single non alphanumeric would be valid as an identifier?

https://learn.microsoft.com/en-us/dotnet/api/system.char.isletterordigit?view=net-8.0

gfs (approved):
One comment. Probably fine either way but something to consider.

src/oss-find-squats-lib/Mutators/IMutator.cs
```diff
@@ -35,6 +36,18 @@ public IEnumerable<Mutation> Generate(PackageURL arg)
             bool hasNamespace = arg.HasNamespace();
             foreach (Mutation mutation in Generate(hasNamespace ? arg.Namespace : arg.Name))
             {
+                if (mutation.Mutated.Length == 0)
+                {
+                    // Don't make mutations that are empty. i.e pkg:npm/ isn't valid.
+                    continue;
+                }
+                
+                if (mutation.Mutated.Length == 1 && !char.IsLetterOrDigit(Convert.ToChar(mutation.Mutated)))
```
gfs, line 45:
```suggestion
                if (mutation.Mutated.Length == 1 && !char.IsLetterOrDigit(
mutation.Mutated[0]))
```
Do we need to convert or just check the first (and only) character of the string?

gfs (approved):
One new comment on using Convert.ToChar vs just taking the first character via indexing. I'm not totally familiar with Convert.ToChar

src/oss-find-squats-lib/Mutators/IMutator.cs
jpinz: Done!

gfs (approved):
LGTM

Issue #406, opened 2023-03-09
GITHUB_ACCESS_TOKEN environment variable is never used


radolin:
Even though it's set in the environment, I always get this error when trying to run oss-health against something hosted in github:

WARN  - Unable to calculate health for pkg:github/stevemao/left-pad: Missing GitHub access token. Please define GITHUB_ACCESS_TOKEN.
I think I have found the reason and will submit a PR shortly, but may be the wrong way to fix it, as I am new to c# or this project.

PR #407, opened 2023-03-09
Fix environment variable processing. (#406)

radolin:
Search for static and non public members as well when overriding values. Here's how the member is defined in GitHubHealthAlgorithm.cs:
private static string? ENV_GITHUB_ACCESS_TOKEN = null;

Searching for public members only misses it and it's never set from the environment variable.
I don't know if this is the right way to fix this, but at least it get's me one step further and I can actually run the oss-health tool for an npm project.

### DIFF
--- a/src/Shared/DefaultHttpClientFactory.cs
+++ b/src/Shared/DefaultHttpClientFactory.cs
@@ -12,7 +12,7 @@ namespace Microsoft.CST.OpenSource
     public sealed class DefaultHttpClientFactory : IHttpClientFactory
     {
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0044:Add readonly modifier", Justification = "Modified through reflection.")]
-        private string ENV_HTTPCLIENT_USER_AGENT = "microsoft_oss_gadget (https://github.com/microsoft/OSSGadget)";
+        public string ENV_HTTPCLIENT_USER_AGENT = "microsoft_oss_gadget (https://github.com/microsoft/OSSGadget)";
 
         public DefaultHttpClientFactory(string? userAgent = null)
         {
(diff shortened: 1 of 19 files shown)


gfs:
Hi @radolin,

Thanks for the report and proposed fix. I'll have to think about this a bit. It might be better just to make the GitHub token field public and non-static - I don't think there's a particular reason for it to be private - or static - but overriding private vars in other circumstances might not be something we want.

gfs:
This report revealed a number of environment variables with the incorrect settings (private or static). I went through and instead of modifying environment helper, changed all properties we intend to be overridden by Environment Helper to be Public instance variables.

radolin:
After your changes I was back to always getting the "Missing GitHub access token. Please define GITHUB_ACCESS_TOKEN." error. I've added 2 commits making ENV_GITHUB_ACCESS_TOKEN a property and changed EnvironmentHelper to look for properties instead of fields. Now it can process the GITHUB_ACCESS_TOKEN environment variable again.

gfs:
Thanks @radolin for catching the needed change to grab properties.

gfs:
@radolin one more thing, can you reply to the Contributor agreement bot? I can't accept the contribution otherwise.

@jpinz any comments/concerns here?

radolin:
> @radolin one more thing, can you reply to the Contributor agreement bot? I can't accept the contribution otherwise.

Will, do, just need to get the blessing from legal for a new CLA. Might take a couple of days.

jpinz (approved):
Assuming it works and everything, LGTM

gfs:
@radolin any update from your legal team?

radolin:
@microsoft-github-policy-service agree company="SUSE"

radolin:
All checks seem green now, sorry for the delay.

PR #408, opened 2023-03-14
Fix issues with PackageVersionExistsAsync for Maven.

jpinz:
Added tests for it, and tests for EnumerateVersions in Maven.

### DIFF
--- a/src/Shared/PackageManagers/MavenProjectManager.cs
+++ b/src/Shared/PackageManagers/MavenProjectManager.cs
@@ -2,7 +2,6 @@
 
 namespace Microsoft.CST.OpenSource.PackageManagers
 {
-    using AngleSharp.Dom;
     using AngleSharp.Html.Parser;
     using Helpers;
     using Microsoft.CST.OpenSource.Contracts;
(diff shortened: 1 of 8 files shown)


src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -166,20 +164,28 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
...
+                foreach (string htmlEntry in htmlEntries)
                 {
-                    if (versionObject != null)
+                    // Get the version.
+                    if (htmlEntry.EndsWith('/') && !htmlEntry.Equals("../"))
                     {
-                        Logger.Debug("Identified {0} version {1}.", packageName, versionObject.InnerText);
-                        versionList.Add(versionObject.InnerText);
+                        var versionStr = htmlEntry.TrimEnd(htmlEntry[^1]);
+                        Logger.Debug("Identified {0} version {1}.", packageName, versionStr);
```
brbayes-msft, line 187: Does this logger statement need to be here?

brbayes-msft (approved):

src/Shared/PackageManagers/MavenProjectManager.cs
jpinz: I guess not? But it's there from the previous code, so I just kept it. We don't see it in our logs I don't think.

gfs (approved):

PR #413, opened 2023-04-14
Parse nuget version to exclude metadata when creating package URL's

morended:
"PackageVersionExistsAsync" and "GetArtifactDownloadUris"  methods return incorrect response when a package version with metadata is requested. This happens as the version string is not normalized as per NuGet standards when creating a download URL. Fixed the logic to parse nuget version using NuGet SDK.

### DIFF
--- a/src/Shared/PackageManagers/NuGetProjectManager.cs
+++ b/src/Shared/PackageManagers/NuGetProjectManager.cs
@@ -63,15 +63,8 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                     $"NuGet package URLs having a repository URL other than '{NUGET_DEFAULT_INDEX}' are not currently supported.");
             }
             
-            var basePath = NUGET_DEFAULT_CONTENT_ENDPOINT.EnsureTrailingSlash();
-            var nameLowercase = purl.Name.ToLowerInvariant();
-            var versionLowercase = NuGetVersion.Parse(purl.Version.ToLowerInvariant());
-
-            yield return new ArtifactUri<NuGetArtifactType>(NuGetArtifactType.Nupkg,
-                $"{basePath}{nameLowercase}/{versionLowercase}/{nameLowercase}.{versionLowercase}.nupkg");
-
-            yield return new ArtifactUri<NuGetArtifactType>(NuGetArtifactType.Nuspec, 
-                $"{basePath}{nameLowercase}/{versionLowercase}/{nameLowercase}.nuspec");
+            yield return new ArtifactUri<NuGetArtifactType>(NuGetArtifactType.Nupkg, GetNupkgUrl(purl.Name, purl.Version));
+            yield return new ArtifactUri<NuGetArtifactType>(NuGetArtifactType.Nuspec, GetNuspecUrl(purl.Name, purl.Version));
         }
 
         /// <inheritdoc />
@@ -236,12 +229,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                 throw new ArgumentNullException(nameof(purl), "Provided PackageURL was null.");
             }
 
-            HttpClient httpClient = CreateHttpClient();
-            
-            // NuGet packages are case insensitive
-            string endpoint = $"{RegistrationEndpoint}{purl.Name.ToLowerInvariant()}/index.json";
-
-            return await CheckHttpCacheForPackage(httpClient, endpoint, useCache);
+            return await this.Actions.DoesPackageExistAsync(purl, useCache);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


morended:
> @microsoft-github-policy-service agree company="Microsoft"

jpinz (approved):
LGTM!

jpinz:
> > @microsoft-github-policy-service agree company="Microsoft"

looks like you quoted it, not sure if that will work

morended:
@microsoft-github-policy-service agree company="Microsoft"

src/oss-tests/ProjectManagerTests/NuGetProjectManagerTests.cs
pmalmsten, line 1: For the time being, Terrapin relies on OSSGadget continuing to tolerate version strings that do not have metadata in them (since Terrapin trims metadata from the package URLs before passing them to OSSGadget today). Could we add test cases here demonstrating that such trimmed package URLs continue to work for package versions that do in fact have metadata in the true version strings?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -263,7 +263,7 @@ public override async Task<bool> PackageVersionExistsAsync(PackageURL purl, bool
             HttpClient httpClient = CreateHttpClient();
 
             // NuGet packages are case insensitive
-            string endpoint = $"{NUGET_DEFAULT_CONTENT_ENDPOINT.EnsureTrailingSlash()}{purl.Name.ToLowerInvariant()}/{purl.Version.ToLowerInvariant()}/{purl.Name.ToLowerInvariant()}.nuspec";
+            string endpoint = $"{NUGET_DEFAULT_CONTENT_ENDPOINT.EnsureTrailingSlash()}{purl.Name.ToLowerInvariant()}/{nugetVersion}/{purl.Name.ToLowerInvariant()}.nuspec";
```
pmalmsten, line 266: Note that we have `this.Actions.DoesPackageExistAsync()` available to us here, which would allow us to defer to the NuGet SDK instead of maintaining this HTTP lookup ourselves. Would we be able to re-write this method to use that instead?

src/Shared/PackageManagers/NuGetProjectManager.cs
```diff
@@ -65,7 +65,7 @@ public override async IAsyncEnumerable<ArtifactUri<NuGetArtifactType>> GetArtifa
             
             var basePath = NUGET_DEFAULT_CONTENT_ENDPOINT.EnsureTrailingSlash();
             var nameLowercase = purl.Name.ToLowerInvariant();
-            var versionLowercase = purl.Version.ToLowerInvariant();
+            var versionLowercase = NuGetVersion.Parse(purl.Version.ToLowerInvariant());
 
             yield return new ArtifactUri<NuGetArtifactType>(NuGetArtifactType.Nupkg,
                 $"{basePath}{nameLowercase}/{versionLowercase}/{nameLowercase}.{versionLowercase}.nupkg");
```
pmalmsten, line 71: There is a helper method called `GetNupkgUrl` available - should we use it?

```suggestion
                GetNupkgUrl(purl.Name, purl.Version);
```

src/oss-tests/ProjectManagerTests/NuGetProjectManagerTests.cs
pmalmsten, line 1: Note that there are methods defined on `BaseProjectManager` that are inherited by `NuGetProjecctManager` - to be truly confident that all methods available on `NuGetProjectManager` work correctly, we should ensure that we should update the tests for all of the public methods offered by `NuGetProjectManager`.

With my prior comment in mind, we should test all of the other methods against both a correct package URL (one which includes metadata in the version string) and a Terrapin trimmed package URL (a package URL for a package version that has metadata in the real package version, but the package URL we give to OSSGadget does not include it).

pmalmsten (changes requested):

gfs (approved):
Looks like a great little fix.

src/oss-tests/ProjectManagerTests/NuGetProjectManagerTests.cs
pmalmsten: Looks good, thanks!

src/Shared/PackageManagers/NuGetProjectManager.cs
pmalmsten: Looks much better, thanks!

src/Shared/PackageManagers/NuGetProjectManager.cs
pmalmsten: This and the new helper method look good.

There's some duplication between `GetNupkgUrl` and the new `GetNuspecUrl` that could probably be factored out, but this is fine as-is.

src/oss-tests/ProjectManagerTests/NuGetProjectManagerTests.cs
pmalmsten: This looks much better, thanks!

pmalmsten (approved):

PR #427, opened 2023-05-30
Updated packageurl-dotnet library.

morended:
I have updated the package-url dotnet library to handle the special characters in the purl components (name, namespace, version, subpath and qualifiers) as per the PURL spec. As part of that change, the PackageUrl now accepts unencoded PURL components and it takes care of encoding/unencoding them when parsing/generating the PURL.  Here is the[ PR](https://github.com/package-url/packageurl-dotnet/pull/23) that has these changes.

I have updated the OSSGadget to use the new dotnet PURL library, as per the new updates, PackageURL constructor now accepts unencoded Purl components as arguments.

### DIFF
--- a/src/oss-find-squats-lib/Mutators/IMutator.cs
+++ b/src/oss-find-squats-lib/Mutators/IMutator.cs
@@ -51,7 +51,7 @@ namespace Microsoft.CST.OpenSource.FindSquats.Mutators
                 if (hasNamespace)
                 {
                     yield return new Mutation(
-                        mutated: arg.CreateWithNewNames(arg.Name, HttpUtility.UrlEncode(mutation.Mutated)).ToString(),
+                        mutated: arg.CreateWithNewNames(arg.Name, mutation.Mutated).ToString(),
                         original: arg.ToString(),
                         reason: mutation.Reason,
                         mutator: mutation.Mutator);
@@ -59,7 +59,7 @@ namespace Microsoft.CST.OpenSource.FindSquats.Mutators
                 else
                 {
                     yield return new Mutation(
-                        mutated: arg.CreateWithNewNames(HttpUtility.UrlEncode(mutation.Mutated), arg.Namespace).ToString(),
+                        mutated: arg.CreateWithNewNames(mutation.Mutated, arg.Namespace).ToString(),
                         original: arg.ToString(),
                         reason: mutation.Reason,
                         mutator: mutation.Mutator);
--- a/src/Shared/Shared.Lib.csproj
+++ b/src/Shared/Shared.Lib.csproj
@@ -45,7 +45,7 @@
         <PackageReference Include="NuGet.Packaging" Version="6.3.1" />
         <PackageReference Include="NuGet.Protocol" Version="6.3.1" />
         <PackageReference Include="Octokit" Version="4.0.1" />
-        <PackageReference Include="packageurl-dotnet" Version="1.2.0" />
+        <PackageReference Include="packageurl-dotnet" Version="1.3.0" />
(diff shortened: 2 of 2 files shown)


jpinz (approved):

src/oss-find-squats-lib/Mutators/IMutator.cs
```diff
@@ -51,15 +51,15 @@ public IEnumerable<Mutation> Generate(PackageURL arg)
                 if (hasNamespace)
                 {
                     yield return new Mutation(
-                        mutated: arg.CreateWithNewNames(arg.Name, HttpUtility.UrlEncode(mutation.Mutated)).ToString(),
+                        mutated: arg.CreateWithNewNames(arg.Name, mutation.Mutated).ToString(),
```
gfs, line 54: Is it safe to remove this URL encoding here? I assume we used to have this to ensure no unescaped potentail dangerous mutations made it out of this method before. What is the rationale for changing it now with this PR? Is there a change inthe packageurl project that is related to if this needs to be preencoded?

pmalmsten: @jpinz[ it looks like you authored the change introducing the original encoding calls](https://github.com/microsoft/OSSGadget/commit/96ac5f107a2d729bee7cd820bbdf353ee0ae5cb3) - would you be able to comment on this? Were these calls originally introduced to work around issues generating correctly-encoded package URLs?

@gfs Mounika has been doing some work upstream in the packageurl-dotnet library to improve how it handles special characters in package names, package versions, and other metadata. Specifically, the packageurl-dotnet library did not implement proper percent- encoding/decoding of special Package URL characters present in attributes of package versions (including `namespace`, `name`, `version`, etc.) as required by spec. The newer version of the packageurl-dotnet library brings the packageurl-dotnet library into compliance with the spec.

As a result, it is now correct to treat individual package URL components handled by the `PackageURL` class (e.g. `Namespace`, `Name`, `Version`, etc.) as always unencoded - the `PackageURL` class takes care of percent- encoding/decoding each component when generating/parsing package URLs. Users of the library no longer need to (and in fact _must_ no longer) pass around individual components of a package URL in encoded form. This makes sense because data encoding is always context-dependent - passing data around in encoded form away from the context in which the encoding matters just makes everyone's lives harder.

Does that help?

@morended I noticed that no description was submitted with this PR. As a general rule, it's wise to include a description with every PR, particularly when bumping a library version requires changing how certain data values are represented. It's important to empathize with code reviewers and give them as much up-front context as we can to help them understand why the changes we propose are appropriate.

morended: Apologies for not updating the description. Updated the description. @pmalmsten  Thanks for clarifying that. 
@jpinz  Let me know if you have any objections with this change.

jpinz: > @jpinz[ it looks like you authored the change introducing the original encoding calls](https://github.com/microsoft/OSSGadget/commit/96ac5f107a2d729bee7cd820bbdf353ee0ae5cb3) - would you be able to comment on this? Were these calls originally introduced to work around issues generating correctly-encoded package URLs?

I don't recall exactly, but I'm pretty sure I added the encoding just to prevent any problems that might occur if they weren't encoded.

gfs: @pmalmsten it sounds like the sanitization is now handled in the dependency so I think this would be fine - but agree with the feedback about including an explanation for the rationale in the description.

gfs (approved):

PR #434, opened 2023-06-28
Escape @s in namespaces in OSS-Health Tool

gfs:
Small refactor to move the escape method to a helper method on the OSSGadget base class so ti can be called from any CLI as needed.

### DIFF
--- a/src/Shared.CLI/OSSGadget.cs
+++ b/src/Shared.CLI/OSSGadget.cs
@@ -11,6 +11,7 @@ namespace Microsoft.CST.OpenSource
     using System.IO;
     using System.Linq;
     using System.Reflection;
+    using System.Text.RegularExpressions;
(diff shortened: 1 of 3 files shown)


pmalmsten (approved):

pmalmsten:
I wonder if there is a more central place that this transform could be done across all CLI tools? I'm not very familiar with how CLI argument parsing is done for OSSGadget - I wonder if there is any place there where we might be able to model a Package URL as a type of value (along the lines of integer vs string vs bool etc.), and put this transform there, so that any CLI tool that defines a Package URL type input value would get this parsing behavior automatically?

gfs:
@pmalmsten I think that would be a better approach, but I'm not sure there's an obvious place to put that. It should be particularly possible for anything that takes just a Purl but for any clis that take multiple arguments it becomes more difficult.

PR #436, opened 2023-06-28
Improved "GetPublishedAtUtcAsync" method efficiency

morended:
Current implementation of "GetPublishedAtUtcAsync" is fetching all the metadata for a given package version, which is causing significant delay. To improve the latency, optimized "GetPublishedAtUtcAsync" method to fetch the uploadtime alone.

### DIFF
--- a/src/Shared/PackageManagers/BaseProjectManager.cs
+++ b/src/Shared/PackageManagers/BaseProjectManager.cs
@@ -18,6 +18,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
     using Utilities;
     using Version = SemanticVersioning.Version;
     using PackageUrl;
+    using System.IO;
 
     public abstract class BaseProjectManager : IBaseProjectManager
     {
@@ -230,13 +231,14 @@ namespace Microsoft.CST.OpenSource.PackageManagers
             Logger.Trace("Loading Uri...");
             HttpResponseMessage result = await client.GetAsync(uri);
             result.EnsureSuccessStatusCode(); // Don't cache error codes.
-            long contentLength = result.Content.Headers.ContentLength ?? 8192;
+            long contentLength = 0;
             JsonDocument doc;
 
             switch (jsonParsingOption)
             {
                 case JsonParsingOption.NotInArrayNotCsv:
                     string data = await result.Content.ReadAsStringAsync();
+                    contentLength = data.Length;
                     data = Regex.Replace(data, @"\r\n?|\n", ",");
                     data = $"[{data}]";
 
@@ -245,8 +247,10 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                         AllowTrailingCommas = true,
                     });
                     break;
-                default: 
-                    doc = await JsonDocument.ParseAsync(await result.Content.ReadAsStreamAsync());
+                default:
+                    Stream responseStream = await result.Content.ReadAsStreamAsync();
+                    contentLength = responseStream.Length;
+                    doc = await JsonDocument.ParseAsync(responseStream);
                     break;
             }
             
@@ -254,7 +258,12 @@ namespace Microsoft.CST.OpenSource.PackageManagers
             {
                 lock (DataCache)
                 {
-                    MemoryCacheEntryOptions? mce = new() { Size = contentLength };
+                    MemoryCacheEntryOptions? mce = new() 
+                    {
+                        Size = contentLength,
+                        SlidingExpiration = TimeSpan.FromMinutes(30),
+                        AbsoluteExpiration = DateTimeOffset.UtcNow.AddHours(6),
+                    };
                     DataCache.Set<JsonDocument>(uri, doc, mce);
                 }
             }
@@ -449,7 +458,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         protected static readonly MemoryCache DataCache = new(
             new MemoryCacheOptions
             {
-                SizeLimit = 1024 * 1024 * 8
+                SizeLimit = 1024 * 1024 * 800
             }
         );
 
--- a/src/Shared/PackageManagers/BaseProjectManager.cs
+++ b/src/Shared/PackageManagers/BaseProjectManager.cs
@@ -436,7 +436,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         }
 
         /// <inheritdoc />
-        public async Task<DateTime?> GetPublishedAtUtcAsync(PackageURL purl, bool useCache = true)
+        public virtual async Task<DateTime?> GetPublishedAtUtcAsync(PackageURL purl, bool useCache = true)
         {
             Check.NotNull(nameof(purl.Version), purl.Version);
             DateTime? uploadTime = (await GetPackageMetadataAsync(purl, useCache))?.UploadTime?.ToUniversalTime();
--- a/src/Shared/PackageManagers/BaseProjectManager.cs
+++ b/src/Shared/PackageManagers/BaseProjectManager.cs
@@ -261,8 +261,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                     MemoryCacheEntryOptions? mce = new() 
                     {
                         Size = contentLength,
-                        SlidingExpiration = TimeSpan.FromMinutes(30),
-                        AbsoluteExpiration = DateTimeOffset.UtcNow.AddHours(6),
+                        AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(30),
                     };
                     DataCache.Set<JsonDocument>(uri, doc, mce);
                 }
@@ -458,7 +457,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         protected static readonly MemoryCache DataCache = new(
             new MemoryCacheOptions
             {
-                SizeLimit = 1024 * 1024 * 800
+                SizeLimit = 1024 * 1024 * 100
             }
         );
 
--- a/src/Shared/PackageManagers/NPMProjectManager.cs
+++ b/src/Shared/PackageManagers/NPMProjectManager.cs
@@ -314,15 +314,7 @@ namespace Microsoft.CST.OpenSource.PackageManagers
             {
                 Version versionToGet = new(metadata.PackageVersion);
                 JsonElement? versionElement = GetVersionElement(contentJSON, versionToGet);
-                
-                if (root.TryGetProperty("time", out JsonElement time))
-                {
-                    string? uploadTime = OssUtilities.GetJSONPropertyStringIfExists(time, metadata.PackageVersion);
-                    if (uploadTime != null)
-                    {
-                        metadata.UploadTime = DateTime.Parse(uploadTime);
-                    }
-                }
+                metadata.UploadTime = ParseUploadTime(contentJSON, metadata.PackageVersion);
 
                 if (versionElement != null)
                 {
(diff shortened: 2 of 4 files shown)


gfs (approved):

pmalmsten (changes requested):
This is generally straightforward - it gets straight to the point, and should run very fast. I'm curious whether you have measured the speed of this change before and after for a package having many thousands of versions, like say https://www.npmjs.com/package/@graphql-codegen/cli?

As a side note, I would point out that the implementations of `GetPackageMetadataAsync` are left without any improvement with this approach. That's not necesssarily a problem, but there is a certain amount of elegance in the approach taken by the prior implentation by leaning on `GetPackageMetadataAsync` to be the one thing that parses package metadata (whereas this approach duplicates some of that parsing logic a bit). That's not a dealbreaker - the approach in this PR simplifies things - but it's a tradeoff that might need to be revisited later.

However, there are other more significant points that we must address when we start thinking about caching:
1. [The memory cache object employed by `GetJsonCache` has a size limit of about 8 MB](https://github.com/microsoft/OSSGadget/blob/d2c21b9f7a7ca26e3289f5e606146d37ab7fe5e6/src/Shared/PackageManagers/BaseProjectManager.cs#L452), whereas the size of the JSON for graphql-codegen/cli is ~25 MB. So we likely need to increase the size limit of the cache, perhaps to say 100 MB. To go with that, we would likely need to bump the memory allocation for Terrapin containers up 50 to 100 additional MB.
2. [When adding entries to the cache, we do not set an expiration time.](https://github.com/microsoft/OSSGadget/blob/d2c21b9f7a7ca26e3289f5e606146d37ab7fe5e6/src/Shared/PackageManagers/BaseProjectManager.cs#L257) As a result, objects may be cached indefinitely. This does not matter all that much for OSS Gadget CLIs that terminate after a few moments, but will matter a lot for Terrapin processes which will keep the running for hours or days at a time. We should probably specify a [sliding expiration](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheentryoptions.slidingexpiration?view=dotnet-plat-ext-7.0#microsoft-extensions-caching-memory-memorycacheentryoptions-slidingexpiration) of something like 30 minutes and an [absolute expiration](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheentryoptions.absoluteexpirationrelativetonow?view=dotnet-plat-ext-7.0#microsoft-extensions-caching-memory-memorycacheentryoptions-absoluteexpirationrelativetonow) of something like 6 hours to ensure that data in Terrapin stays fresh. It would be ideal to do this for all places where we set values in the cache.

Those are what come to mind first - getting caching right can be tricky, so I'll let you know if I think of anything else later. Let me know if you have any questions on the above.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -482,6 +482,23 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
+            JsonDocument jsonDoc = await GetJsonCache(client, $"{ENV_NPM_API_ENDPOINT}/{packageName}", useCache);
+            if (jsonDoc.RootElement.TryGetProperty("time", out JsonElement time))
+            {
+                string? uploadTime = OssUtilities.GetJSONPropertyStringIfExists(time, purl.Version);
+                if (uploadTime != null)
+                {
+                    return DateTime.Parse(uploadTime).ToUniversalTime();
+                }
+            }
+            return null;
```
pmalmsten, line 499: Nit: for this if statement, should we extract it to a helper method so that we don't have to duplicate this logic between this method and `GetPackageMetadata`?

pmalmsten:
We also discussed offline how NPM does not set the `Content-Length` header on HTTP responses (which causes `GetJsonCache` to signifigcantly underestimate the size of objects that are cached), and that we should fix the response size estimation in OSSGadget so that we just count how many bytes are in response bodies instead of trying to use an optional header.

morended:
> This is generally straightforward - it gets straight to the point, and should run very fast. I'm curious whether you have measured the speed of this change before and after for a package having many thousands of versions, like say https://www.npmjs.com/package/@graphql-codegen/cli?
> 
> As a side note, I would point out that the implementations of `GetPackageMetadataAsync` are left without any improvement with this approach. That's not necesssarily a problem, but there is a certain amount of elegance in the approach taken by the prior implentation by leaning on `GetPackageMetadataAsync` to be the one thing that parses package metadata (whereas this approach duplicates some of that parsing logic a bit). That's not a dealbreaker - the approach in this PR simplifies things - but it's a tradeoff that might need to be revisited later.
> 
> However, there are other more significant points that we must address when we start thinking about caching:
> 
> 1. [The memory cache object employed by `GetJsonCache` has a size limit of about 8 MB](https://github.com/microsoft/OSSGadget/blob/d2c21b9f7a7ca26e3289f5e606146d37ab7fe5e6/src/Shared/PackageManagers/BaseProjectManager.cs#L452), whereas the size of the JSON for graphql-codegen/cli is ~25 MB. So we likely need to increase the size limit of the cache, perhaps to say 100 MB. To go with that, we would likely need to bump the memory allocation for Terrapin containers up 50 to 100 additional MB.
> 2. [When adding entries to the cache, we do not set an expiration time.](https://github.com/microsoft/OSSGadget/blob/d2c21b9f7a7ca26e3289f5e606146d37ab7fe5e6/src/Shared/PackageManagers/BaseProjectManager.cs#L257) As a result, objects may be cached indefinitely. This does not matter all that much for OSS Gadget CLIs that terminate after a few moments, but will matter a lot for Terrapin processes which will keep the running for hours or days at a time. We should probably specify a [sliding expiration](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheentryoptions.slidingexpiration?view=dotnet-plat-ext-7.0#microsoft-extensions-caching-memory-memorycacheentryoptions-slidingexpiration) of something like 30 minutes and an [absolute expiration](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheentryoptions.absoluteexpirationrelativetonow?view=dotnet-plat-ext-7.0#microsoft-extensions-caching-memory-memorycacheentryoptions-absoluteexpirationrelativetonow) of something like 6 hours to ensure that data in Terrapin stays fresh. It would be ideal to do this for all places where we set values in the cache.
> 
> Those are what come to mind first - getting caching right can be tricky, so I'll let you know if I think of anything else later. Let me know if you have any questions on the above.

As discussed offline, increased the cache limit and added cache expiration.

morended:
> We also discussed offline how NPM does not set the `Content-Length` header on HTTP responses (which causes `GetJsonCache` to signifigcantly underestimate the size of objects that are cached), and that we should fix the response size estimation in OSSGadget so that we just count how many bytes are in response bodies instead of trying to use an optional header.

I have changed this to get contentlength from the Http response body.

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -230,13 +231,14 @@ public static async Task<JsonDocument> GetJsonCache(HttpClient client, string ur
             Logger.Trace("Loading Uri...");
             HttpResponseMessage result = await client.GetAsync(uri);
             result.EnsureSuccessStatusCode(); // Don't cache error codes.
-            long contentLength = result.Content.Headers.ContentLength ?? 8192;
+            long contentLength = 0;
```
pmalmsten, line 234: Nit: It's actually slightly better not to assign a default value here since we expect any/all code paths below to set the value. And the compiler will in fact help us make sure that all code paths set the value.

```suggestion
            long contentLength;
```

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -245,16 +247,23 @@ public static async Task<JsonDocument> GetJsonCache(HttpClient client, string ur
...
             if (useCache)
             {
                 lock (DataCache)
                 {
-                    MemoryCacheEntryOptions? mce = new() { Size = contentLength };
+                    MemoryCacheEntryOptions? mce = new() 
+                    {
+                        Size = contentLength,
+                        SlidingExpiration = TimeSpan.FromMinutes(30),
+                        AbsoluteExpiration = DateTimeOffset.UtcNow.AddHours(6),
```
pmalmsten, line 265: Nit: there is a built-in way to specify a relative expiration time instead of a strictly absolute expiration time:

```suggestion
                        AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(6),
```

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -449,7 +458,7 @@ public virtual List<Version> GetVersions(JsonDocument? metadata)
         protected static readonly MemoryCache DataCache = new(
             new MemoryCacheOptions
             {
-                SizeLimit = 1024 * 1024 * 8
+                SizeLimit = 1024 * 1024 * 800
```
pmalmsten, line 461: Is the intent here to raise the size limit from 8 MB to 800 MB? Our Docker containers have finite memory allocations, typically only around 800 MB total - so 800 MB for just cache would be a bit much.

How about 100 MB instead?

```suggestion
                SizeLimit = 1024 * 1024 * 100, // 100 MB
```

src/Shared/PackageManagers/PyPIProjectManager.cs
```diff
@@ -432,6 +432,37 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
...
+                    {
+                        DateTime newUploadTime = DateTime.Parse(uploadTimeStr).ToUniversalTime();
+                        // Used to set the minimum upload time for all associated files for this version to get the publish time.
+                        if (uploadTime == null || uploadTime > newUploadTime)
+                        {
+                            uploadTime = newUploadTime;
+                        }
+                    }
+                }
+            }
```
pmalmsten, line 461: Should we also extract the above logic into a helper method (like what you did for NPM), to avoid duplication?

src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -245,16 +247,23 @@ public static async Task<JsonDocument> GetJsonCache(HttpClient client, string ur
                         AllowTrailingCommas = true,
                     });
                     break;
-                default: 
-                    doc = await JsonDocument.ParseAsync(await result.Content.ReadAsStreamAsync());
+                default:
+                    Stream responseStream = await result.Content.ReadAsStreamAsync();
+                    contentLength = responseStream.Length;
```
pmalmsten, line 252: Huh - well that's pretty convenient. I thought we might need a stream-wrapper approach in order to dynamically calculate the total length of the stream, but it looks like we don't have to do that (at least for now).

pmalmsten (changes requested):
Good progress, but a few things could still use tweaking.

src/Shared/PackageManagers/PyPIProjectManager.cs
morended: I looked at it, the parsing logic to get the upload time involves parsing a json array which has other metadata apart from the upload time. If we use a helper method to move the logic to extract upload time from json array, that logic is duplicated in the GetMetadata method as we need the array parsing to extract other metadata. There is only 3-4 lines of code to get the minimum upload time that can be extracted to helper method. I thought it should be ok to leave it as is.

src/Shared/PackageManagers/BaseProjectManager.cs
morended: Addressed in latest revision

src/Shared/PackageManagers/BaseProjectManager.cs
morended: Addressed in latest version.

src/Shared/PackageManagers/PyPIProjectManager.cs
pmalmsten: OK, that makes sense. Thanks for looking into it.

pmalmsten (approved):
Latest changes look great! Thanks Mounika.

PR #438, opened 2023-07-07
Skip cocoapods package download test.

morended:


gfs (approved):
Approved as workaround until underlying version enumerating behavior for Cocoapods can be rewritten.

PR #439, opened 2023-07-07
Fix cocoapods version enumeration

### LINKED ISSUE #437: Unable to fetch versions for a  Cocoapods package
The cocoapods gets the list of versions by parsing the html for the page where cocoapods specs are stored. Example : https://github.com/CocoaPods/Specs/tree/master/Specs/8/2/7/RandomKit

The html page is changed and hence the query selectors used to fetch version data no more exists on the page, as a result causing it to fail.

gfs:
Fix #437

### DIFF
--- a/src/Shared/PackageManagers/CocoapodsProjectManager.cs
+++ b/src/Shared/PackageManagers/CocoapodsProjectManager.cs
@@ -2,9 +2,11 @@
 
 namespace Microsoft.CST.OpenSource.PackageManagers
 {
+    using AngleSharp.Html.Dom;
     using AngleSharp.Html.Parser;
     using Extensions;
     using Helpers;
+    using Newtonsoft.Json.Linq;
     using PackageUrl;
     using System;
     using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/Shared/PackageManagers/CocoapodsProjectManager.cs
```diff
@@ -158,17 +160,22 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
                 string? html = await GetHttpStringCache(httpClient, $"{ENV_COCOAPODS_SPECS_ENDPOINT}/Specs/{prefix}/{packageName}");
                 HtmlParser parser = new();
                 AngleSharp.Html.Dom.IHtmlDocument document = await parser.ParseDocumentAsync(html);
-                AngleSharp.Dom.IHtmlCollection<AngleSharp.Dom.IElement> navItems = document.QuerySelectorAll("div.Details a.js-navigation-open");
+                // Fetch the embedded react data
+                string innerHtml = document.QuerySelector("script[data-target='react-app.embeddedData']").InnerHtml;
```
morended, line 164: Nit: We might run into the same issue if the html is changed again. Did we consider other options here?

morended (approved):

src/Shared/PackageManagers/CocoapodsProjectManager.cs
gfs: I think we do want to switch to a more permanent solution in the future - I'd recommend a switch to libgit2sharp, we can list the directories in the appropriate location in the specs repository. 

This is just a bandaid fix to get it working again until we have time to do a better fix.

scovetta: Are you sure you can do that remotely? I thought you'd need to clone the entire repo in order to enumerate files. You could use Octocat or another GitHub library but then you'd need the user to supply an access token.

gfs: Fair point - I'm not positive you can do that remotely. I've only used libgit2sharp to clone/checkout and then perform operations in the past.  If we do have to switch to octocat and have a token it adds another layer of complexity as well.

scovetta (approved):

PR #450, opened 2023-12-01
Add line number for excerpt in oss-detect-backdoor

Flying-Tom:
Add line number for excerpt in oss-detect-backdoor (resolves  #449 )

PR #451, opened 2023-12-06
Change the way of excerpts line to be splited

Flying-Tom:
Use Environment.NewLine to replace new[] { "\r", "\n", "\r\n" }. The latter can result in peculiar line breaks in certain code files, such as introducing additional blank lines in Python source code.

### DIFF
--- a/src/oss-detect-backdoor/DetectBackdoorTool.cs
+++ b/src/oss-detect-backdoor/DetectBackdoorTool.cs
@@ -121,7 +121,7 @@ namespace Microsoft.CST.OpenSource
                         Console.WriteLine("  Severity: " + Cyan(match.Severity.ToString()) + ", Confidence: " + Cyan(match.Confidence.ToString()));
                         Console.WriteLine("  Filename: " + Yellow(filename));
                         Console.WriteLine("   Pattern: " + Green(match.MatchingPattern.Pattern));
-                        foreach (string? line in match.Excerpt.Split(new[] {"\r", "\n", "\r\n"}, StringSplitOptions.None))
+                        foreach (string? line in match.Excerpt.Split(Environment.NewLine, StringSplitOptions.None))
                         {
                             string? s = line;
                             if (s.Length > 100)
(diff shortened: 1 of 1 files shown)


gfs:
Hi @flying-tom thanks for your feedback. Can you give an example of a python file where you're seeing undesirable behavior?

We can't take this change as is, the choice to use the three line separators is intentional, using `Environment.NewLine` makes line splitting platform dependent, so you'll end up with different behavior if you run on Linux or Windows, but the source code being scanned didn't necessarily come from the same platform. This results in a number of undesirable behaviors - for example if you split lines for a file that used `\n` on windows you end up with just one line because there are no carriage returns to match, and if you split lines from a windows file with carriage returns on linux you end up with trailing carriage returns at the end of each line. 

The extra blank lines may be down to the order of the newline characters in the split, perhaps the `\r\n` has to be first.

gfs:
It looks like the ordering of the split tokens is the culprit.
image

gfs:
And examples of the undesirable behavior when using Environment.NewLine (with emulated behavior for nix execution)
image

Flying-Tom:
> Hi @Flying-Tom thanks for your feedback. Can you give an example of a python file where you're seeing undesirable behavior?
> 
> We can't take this change as is, the choice to use the three line separators is intentional, using `Environment.NewLine` makes line splitting platform dependent, so you'll end up with different behavior if you run on Linux or Windows, but the source code being scanned didn't necessarily come from the same platform. This results in a number of undesirable behaviors - for example if you split lines for a file that used `\n` on windows you end up with just one line because there are no carriage returns to match, and if you split lines from a windows file with carriage returns on linux you end up with trailing carriage returns at the end of each line.
> 
> The extra blank lines may be down to the order of the newline characters in the split, perhaps the `\r\n` has to be first.

Indeed, it was my oversight （This fix works in my case so i didn't think about the design intention)

Flying-Tom:
> It looks like the ordering of the split tokens is the culprit. image

I believe this is an appropriate fix.

gfs (approved):
LGTM. Thanks for the fix!

gfs:
It seems like you've incorporated this into the previous PR #450, so closing this as now unneeded since the previous was merged. If I've misinterpreted feel free to comment and reopen.

PR #460, opened 2024-03-18
Add method to NpmPackageManager to fime the time at which the first package version is created.

morended:


src/Shared/PackageManagers/BaseProjectManager.cs
```diff
@@ -411,6 +411,13 @@ public virtual List<Version> GetVersions(JsonDocument? metadata)
             throw new NotImplementedException($"{typeName} does not implement GetVersions.");
         }
 
+        /// <inheritdoc />
+        public virtual Task<DateTime?> GetPackageCreatedAtUtcAsync(PackageURL purl, bool useCache = true)
+        {
+            string typeName = GetType().Name;
+            throw new NotImplementedException($"{typeName} does not implement GetPackageCreatedAtUtcAsync.");
+        }
```
pmalmsten, line 419: Should we really add a new interface method & base class implementation if for the vast majority of package managers it just throws an exception? I wonder whether it would make more sense to just implement this on NPMPackageManager for now and to only elevate it to IBaseProjectManager once more package managers add support for it - otherwise it feels like false advertising in a way.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -482,6 +482,26 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
             JsonDocument jsonDoc = await GetJsonCache(client, $"{ENV_NPM_API_ENDPOINT}/{packageName}", useCache);
             return ParseUploadTime(jsonDoc, purl.Version);
         }
+        public override async Task<DateTime?> GetPackageCreatedAtUtcAsync(PackageURL purl, bool useCache = true)
+        {
+            HttpClient client = CreateHttpClient();
+            string? packageName = purl.HasNamespace() ? $"{purl.GetNamespaceFormatted()}/{purl.Name}" : purl.Name;
+            JsonDocument jsonDoc = await GetJsonCache(client, $"{ENV_NPM_API_ENDPOINT}/{packageName}", useCache);
+            return ParseCreatedTime(jsonDoc);
+        }
```
pmalmsten, line 491: It's not something we need to address in this particular PR, but the current separate-method-for-each-separate-metadata-field approach feels a bit clunky and might involve redundant JSON parsing depending on how efficient `GetJsonCache` is.

There are always tradeoffs - but at least for npm where we have to parse a bunch of stuff to get a single property anyway we might as well put all of the stuff we parsed into a single 'metadata' object in one go and return that one object (instead of making separate NPMProjectManager method calls for every individual thing we might want)

gfs: I agree it's different conceptually than this PR, but agree that a structured metadata object would be preferable to adding new methods for each field.

src/Shared/PackageManagers/BaseProjectManager.cs
morended: I have removed the method in favor of using GetPackageMetadataAsync method to fetch metadata.

src/Shared/PackageManagers/NPMProjectManager.cs
morended: I have removed the separate method to fetch the individual metadata fields in favor of using GetPackageMetadataAsync method to fetch metadata.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -278,6 +278,11 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
         /// <inheritdoc />
         /// <remarks>Currently doesn't respect the <paramref name="includePrerelease"/> flag.</remarks>
         public override async Task<PackageMetadata?> GetPackageMetadataAsync(PackageURL purl, bool includePrerelease = false, bool useCache = true)
+        {
+           return await GetNpmPackageMetadataAsync(purl, includePrerelease, useCache);
+        }
+
+        public  async Task<PackageMetadata?> GetNpmPackageMetadataAsync(PackageURL purl, bool includePrerelease = false, bool useCache = true, bool includeRepositoryMetadata = true)
```
pmalmsten, line 285: Tiniest nit: I believe in C# that optional parameters having a default value provided work the same as having an additional overload which does not specify the argument at all - so the overload on lines 281-283 may not be needed.

pmalmsten (approved):

src/Shared/PackageManagers/NPMProjectManager.cs
morended: The first method is override of GetPackageMetadataAsync() method used by all package managers. The second method is specific to NpmPackageManager where we want an option to skip collecting repository metadata. Had we only used the second method with default parameter, the existing code using the BaseProjectManager object to fetch the package metadata from "GetPackageMetadataAsync" method will be impacted as the new method is specific to NpmProjectManager,

pmalmsten: @morended You're right, I did not notice that they are different methods. I admit I am confused though - why are they separate? (If it's because `GetPackageMetadataAsync` implements an interface, it would seem pretty reasonable to put the new optional argument on the existing interface)

morended: I thought this argument is Npm specific and all other package Mangers might not need to implement this.

pmalmsten: @morended When we were adding a new method that was only implemented for npm, that was definitely npm-specific. 

But making a change to an existing interface implementation is significantly less npm-specific. Also I could see this new flag being honored by other implementations as well - if the npm package manager fetches repository info from GitHub, it seems plausible that other BasePackageManager implementations might do the same thing now or in the future.

pmalmsten (approved):
Looks great! 👍

gfs (approved):

PR #461, opened 2024-03-21
Ignore cargo test failing in pipeline.

morended:


gfs (approved):
Confirmed that test is succeeding locally but failing in pipeline. Appears to be an issue with mocking that only seems to reproduce in CI.

PR #475, opened 2024-10-25
Add optional timeouts for project managers

gregorymacharia:
On a previous incident occurring due to a disruption in npmjs.org, there were very long response times (~60s) that ultimately came back with 503 responses. This impacted Terrapin's latencies and availability. This change allows users to specify timeouts when they use OSSGadget's project managers when they call `CreateProjectManager`

### DIFF
--- a/src/Shared/PackageManagers/NPMProjectManager.cs
+++ b/src/Shared/PackageManagers/NPMProjectManager.cs
@@ -130,6 +130,11 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                     downloadedPaths.Add(extractionPath);
                 }
             }
+            catch (TaskCanceledException ex)
+            {
+                Logger.Debug(ex, "Downloading NPM package timed out: {0}", ex.Message);
+                throw;
+            }
             catch (Exception ex)
             {
                 Logger.Debug(ex, "Error downloading NPM package: {0}", ex.Message);
--- a/src/Shared/PackageManagers/NPMProjectManager.cs
+++ b/src/Shared/PackageManagers/NPMProjectManager.cs
@@ -43,8 +43,9 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         public NPMProjectManager(
             string directory,
             IManagerPackageActions<IManagerPackageVersionMetadata>? actions = null,
-            IHttpClientFactory? httpClientFactory = null)
-            : base(actions ?? new NoOpPackageActions(), httpClientFactory ?? new DefaultHttpClientFactory(), directory)
+            IHttpClientFactory? httpClientFactory = null,
+            TimeSpan? timeout = null)
+            : base(actions ?? new NoOpPackageActions(), httpClientFactory ?? new DefaultHttpClientFactory(), directory, timeout)
         {
         }
 
--- a/src/oss-tests/ProjectManagerTests/ProjectManagerFactoryTests.cs
+++ b/src/oss-tests/ProjectManagerTests/ProjectManagerFactoryTests.cs
@@ -3,13 +3,16 @@
 namespace Microsoft.CST.OpenSource.Tests.ProjectManagerTests;
 
 using Contracts;
+using Moq;
 using PackageActions;
 using PackageManagers;
 using PackageUrl;
+using RichardSzalay.MockHttp;
 using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Threading.Tasks;
 using VisualStudio.TestTools.UnitTesting;
 
 [TestClass]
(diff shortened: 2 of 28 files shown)


gfs (approved):
Seems like a fine change to make, though not clear to me that the tests here are actually exercising the timeout functionality itself to confirm its working which should be possible by mocking the request as a sleep.

gregorymacharia:
> Seems like a fine change to make, though not clear to me that the tests here are actually exercising the timeout functionality itself to confirm its working which should be possible by mocking the request as a sleep.

Makes sense I will add another commit to mock a delay and add to those tests

gfs (approved):
With the test to confirm timeout functionality I think this LGTM.

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -129,6 +130,11 @@ public override async Task<IEnumerable<string>> DownloadVersionAsync(PackageURL
                     downloadedPaths.Add(extractionPath);
                 }
             }
+            catch (TaskCanceledException ex)
```
pmalmsten, line 133: According to the remarks on HttpClient.Timeout, it throws WebException when the timeout expires: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=net-8.0#remarks How do we know that TaskCanceledException indicates a timeout occurred? Should we also catch WebException (or catch it instead of TaskCanceledException)?

src/oss-tests/ProjectManagerTests/ProjectManagerFactoryTests.cs
```diff
@@ -127,16 +130,70 @@ public void RemoveAllProjectManagersSucceeds()
...
+    public async Task PackageManagerRequestsTimeOutCorrectly()
+    {
+        // Arrange
+        Mock<IHttpClientFactory> mockFactory = new();
+        MockHttpMessageHandler mockHttp = new();
+        mockHttp
+            .When(HttpMethod.Get, "*")
+            .Respond(async () =>
+            {
+                await Task.Delay(5000); // simulate a 5 seconds delay
```
pmalmsten, line 176: Whenever we delay using wall-clock time in unit tests, that can be pretty annoying (since developers sit through the delay whenever the tests run). (The best way to check timeouts in unit tests would be to simulate the passage of time by mocking a time provider interface, so that unit tests don't actually wait at all - but since mockHTTP is integrating with built-in .NET HTTP classes, that doesn't seem to be an option unfortunately).

I see that the timeout you specified is much smaller (100 ms); could this delay be 120ms?

src/Shared/PackageManagers/NPMProjectManager.cs
gregorymacharia: Not sure about the discrepancy from the documentation but from testing and looking at the HttpClient Code, it throws a TimeoutException wrapped in a TaskCanceledException. I have added a check to assert that the inner exception is of type TimeoutException.

src/oss-tests/ProjectManagerTests/ProjectManagerFactoryTests.cs
gregorymacharia: Yes, good catch I meant to have that as a smaller delay here. Updated it to be 120ms.

src/Shared/PackageManagers/NPMProjectManager.cs
pmalmsten: Would you mind trying quickly to confirm? Something like in a unit test, temporarily add a couple lines of code to make a HTTP request (using the real thing, not a mock), set the timeout to be ridiculously low (like 1ms), and try sending a request to google.com or something.

src/oss-tests/ProjectManagerTests/ProjectManagerFactoryTests.cs
pmalmsten: Looks great, thanks!

src/Shared/PackageManagers/NPMProjectManager.cs
gregorymacharia: > Would you mind trying quickly to confirm? Something like in a unit test, temporarily add a couple lines of code to make a HTTP request (using the real thing, not a mock), set the timeout to be ridiculously low (like 1ms), and try sending a request to google.com or something.

Yes. See below - The second test fails with message: 
"Assert.IsInstanceOfType failed.  Expected type:<System.Net.WebException>. Actual type:<System.Threading.Tasks.TaskCanceledException>."

image
image

pmalmsten: Great thanks for confirming!

pmalmsten (approved):

PR #483, opened 2024-12-02
fix path for linux/macos signing

danfiedler-msft:


danfiedler-msft:
replaced with https://github.com/microsoft/OSSGadget/pull/484

PR #493, opened 2025-03-07
Update cargo package manager to fetch cargo metadata from non-rate limited API

morended:
Updated Cargo Package Manager to use non rate limited API as per https://crates.io/data-access.

- Crates does not have a non-ratelimited API to fetch metadata for a given package version. Hence updated the "GetMetadataAsync" method to throw "NotImplementedException".
- Updated the download urls to be fetched from "https://static.crates.io/" (non-ratelimited API).


michelleqyun:
Are the not implemented exceptions temporary while we find a new way to retrieve metadata and published timestamp? Does this mean quarantine is disabled for cargo packages?

morended:
> Are the not implemented exceptions temporary while we find a new way to retrieve metadata and published timestamp? Does this mean quarantine is disabled for cargo packages?

This is temporary until we figure out a way to fetch metadata for cargo packages. With these changes, quarantine will be disabled for cargo packages. I am working on a fix for quarantine.

PR #494, opened 2025-03-14
Update Cargo Package Manager to fetch publishedtimestamp and download uris from non-ratelimited API

morended:
- Updated "GetPublishedAtUtcAsync" to use non rate limited API (https://static.crates.io/rss/crates/{name}.xml) to fetch published timestamp of a package version when the request volume is high.

- Updated "GetArtifactDownloadUrisAsync" the download URLs to be fetched from "https://static.crates.io/" (non-rate limited API).


michelleqyun:
The get package metadata method should only be called when rate limit is low, correct? Is there a way to get package metadata when rate limit is high? And is it up to the caller to manage rate limiting (no additional checks in place)?

morended:
> The get package metadata method should only be called when rate limit is low, correct? Is there a way to get package metadata when rate limit is high? And is it up to the caller to manage rate limiting (no additional checks in place)?

There isn't afaik, as crates doesn't provide any non-rate limited API to fetch metadata today.

michelleqyun (approved):

src/Shared/PackageManagers/CargoProjectManager.cs
```diff
@@ -298,6 +304,58 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
...
+                XDocument doc = XDocument.Parse(rssFeedContent);
+                List<XElement> items = doc.Descendants("item").ToList();
+
+                foreach (XElement item in items)
+                {
+                    // Using LocalName to account for namespaces in custom elements
+                    string? pubDate = item.Elements().FirstOrDefault(e => e.Name == "pubDate")?.Value;
+                    string? crateName = item.Elements().FirstOrDefault(e => e.Name.LocalName == "name" && e.Name.NamespaceName == $"{ ENV_CARGO_ENDPOINT}/")?.Value;
+                    string? crateVersion = item.Elements().FirstOrDefault(e => e.Name.NamespaceName == $"{ENV_CARGO_ENDPOINT}/" && e.Name.LocalName == "version")?.Value;
+                    if (crateName == purl.Name && crateVersion == purl.Version)
```
OmikMahajan, line 338: nit: Can we keep a local cache for pairs for (name,version) to their published timestamp? It'll help us avoid parsing the entire XML for every request which relies on using the RSS feed.

src/Shared/PackageManagers/CargoProjectManager.cs
```diff
@@ -298,6 +304,58 @@ public override Uri GetPackageAbsoluteUri(PackageURL purl)
             return new Uri(url);
         }
 
+        public override async Task<DateTime?> GetPublishedAtUtcAsync(PackageURL purl, bool useCache = true, bool highRequestVolume = false)
+        {
+            DateTime? uploadTime = await GetPackageVersionPublishedTimestampFromRssFeed(purl, useCache);
```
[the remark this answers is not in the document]
morended: Yes, If the non-rate-limited API provides a timestamp, we don't have to call the rate-limited API. This approach is intended to minimize the use of the rate-limited API.

src/Shared/PackageManagers/CargoProjectManager.cs
morended: The RSS feed provides updates for each package, which includes versions published within the last 24 hours. Given the small size of the response (approximately 10 package versions), it is currently uncertain whether caching the parsed content is necessary, as the package versions change every 24 hours. Additionally, the HTTP response is already cached to minimize network calls.

src/Shared/PackageManagers/CargoProjectManager.cs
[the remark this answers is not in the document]
morended: They will call the rate-limited API if the non-rate-limited API returns null. This could lead to an additional network call, which should be acceptable as the latency is manageable for local development. Is there any reason for low traffic requests to consistently use the rate-limited API?

src/Shared/PackageManagers/CargoProjectManager.cs
morended: Added a comment to clarify.

src/Shared/PackageManagers/CargoProjectManager.cs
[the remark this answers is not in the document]
OmikMahajan: Given that the RSS feed only contains data for the last 24 hours - I think that checking it before calling the rate-limited API is okay.

OmikMahajan (approved):

src/Shared/Contracts/IBaseProjectManager.cs
```diff
@@ -154,6 +154,7 @@ public interface IBaseProjectManager
     /// </summary>
     /// <param name="purl">Package URL specifying the package. Version is mandatory.</param>
     /// <param name="useCache">If the cache should be used when looking for the published time.</param>
+    /// <param name="highRequestVolume">If the request load is high, use non rate-limited API to fetch timestamp.</param>
     /// <returns>The <see cref="DateTime"/> when this version was published, or null if not found.</returns>
-    public Task<DateTime?> GetPublishedAtUtcAsync(PackageURL purl, bool useCache = true);
+    public Task<DateTime?> GetPublishedAtUtcAsync(PackageURL purl, bool useCache = true, bool highRequestVolume = false);
```
pmalmsten, line 159: Today this high volume status only applies to Cargo. Have you considered making this be a property of the Cargo Project Manager constructor instead of this interface method in order to reduce the scope of these changes to other package managers?

It's not great for an interface to require multiple package manager classes to support this new boolean flag when the flag only does something for one type of package. That suggests to users of this API that the package managers might be able to do something or behave differently in a meaningful way when they cannot.

pmalmsten: That could be plumbed through the ProjectManagerFactory constructor to make it available to the CargoProjectManager.

I know it being a ProjectManagerFactory may still lead to the possibility of user confusion as to which project manager(s) handle this and which don't. But if we document it along the lines of 'For project managers whose implementations may use a rate-limited API, this disables calling any rate-limited API when true' and it might be OK.

pmalmsten, line 159: The naming `highRequestVolume` is ambiguous for an API specification. What is 'high'?

I wonder it would be better to take a different angle, and call this something like `allowUseOfRateLimitedRegistryAPIs = true`. Would that be more direct at describing what this really means?

morended: Makes sense. Made the changes to add the flag to the constructor.

morended: Addressed!

src/Shared/Contracts/IProjectManagerFactory.cs
```diff
@@ -14,6 +14,8 @@ public interface IProjectManagerFactory
     /// <param name="purl">The <see cref="PackageURL"/> for the package to create the project manager for.</param>
     /// <param name="destinationDirectory">The new destination directory, if provided.</param>
     /// <param name="timeout">The <see cref="TimeSpan"/> to wait before the request times out.</param>
+    /// <param name="allowUseOfRateLimitedRegistryAPIs"> The Cargo Project Manager uses this flag to disable the use of a ratelimited API for 
+    /// fetching metadata when the user scenario requires retrieving metadata for more than one package per second. </param>
     /// <returns>The implementation of <see cref="BaseProjectManager"/> for this <paramref name="purl"/>'s type.</returns>
-    IBaseProjectManager? CreateProjectManager(PackageURL purl, string destinationDirectory = ".", TimeSpan? timeout = null);
+    IBaseProjectManager? CreateProjectManager(PackageURL purl, string destinationDirectory = ".", TimeSpan? timeout = null, bool? allowUseOfRateLimitedRegistryAPIs = true);
```
pmalmsten, line 20: I know that we need to vary the timeout value per request (some requests can tolerate longer latency, others not); do we need to vary this API rate limiting setting per request?

If not, I wonder whether this could be a constructor parameter on ProjectManagerFactory, so that it would only need to be specified once for a whole project (like how the HTTP client factory is supplied once via the ProjectManagerFactory constructor)?

morended: Moved the parameter to ProjectManagerFactory constructor to be only used when creating CargoPackageManager

src/Shared/PackageManagers/CargoProjectManager.cs
```diff
@@ -54,9 +57,11 @@ public CargoProjectManager(
             string directory,
             IManagerPackageActions<IManagerPackageVersionMetadata>? actions = null,
             IHttpClientFactory? httpClientFactory = null,
-            TimeSpan? timeout = null)
+            TimeSpan? timeout = null,
+            bool? allowUseOfRateLimitedRegistryAPIs = null)
             : base(actions ?? new NoOpPackageActions(), httpClientFactory ?? new DefaultHttpClientFactory(), directory, timeout)
         {
+            this.allowUseOfRateLimitedRegistryAPIs = allowUseOfRateLimitedRegistryAPIs ?? true;
         }
```
pmalmsten, line 65:
```suggestion
            bool allowUseOfRateLimitedRegistryAPIs = true)
            : base(actions ?? new NoOpPackageActions(), httpClientFactory ?? new DefaultHttpClientFactory(), directory, timeout)
        {
        }
```

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -44,17 +44,19 @@ public ProjectManagerFactory(Dictionary<string, ConstructProjectManager> overrid
         /// Initializes a new instance of a <see cref="ProjectManagerFactory"/>.
         /// </summary>
         /// <param name="httpClientFactory">The <see cref="IHttpClientFactory"/> to use in <see cref="GetDefaultManagers"/>.</param>
-        public ProjectManagerFactory(IHttpClientFactory? httpClientFactory = null)
+        /// <param name="allowUseOfRateLimitedRegistryAPIs"> The Cargo Project Manager uses this flag to disable the use of a ratelimited API for 
+        /// fetching metadata when the user scenario requires retrieving metadata for more than one package per second. </param>
+        public ProjectManagerFactory(IHttpClientFactory? httpClientFactory = null, bool? allowUseOfRateLimitedRegistryAPIs = null)
```
pmalmsten, line 49:
```suggestion
        public ProjectManagerFactory(IHttpClientFactory? httpClientFactory = null, bool allowUseOfRateLimitedRegistryAPIs = true)
```

src/Shared/PackageManagers/ProjectManagerFactory.cs
```diff
@@ -44,17 +44,19 @@ public ProjectManagerFactory(Dictionary<string, ConstructProjectManager> overrid
...
+            _projectManagers = GetDefaultManagers(httpClientFactory, allowUseOfRateLimitedRegistryAPIs);
         }
 
         /// <summary>
         /// Constructs the default set of ProjectManagers (optionally using a specified <see cref="IHttpClientFactory"/>) that ship with OSSGadget.
         /// </summary>
         /// <param name="httpClientFactoryParam">The <see cref="IHttpClientFactory"/> to use in the project managers.</param>
         /// <returns>A dictionary of the default managers with associated constructors.</returns>
-        public static Dictionary<string, ConstructProjectManager> GetDefaultManagers(IHttpClientFactory? httpClientFactoryParam = null)
+        public static Dictionary<string, ConstructProjectManager> GetDefaultManagers(IHttpClientFactory? httpClientFactoryParam = null, bool? allowUseOfRateLimitedRegistryAPIs = null)
```
pmalmsten, line 59:
```suggestion
        public static Dictionary<string, ConstructProjectManager> GetDefaultManagers(IHttpClientFactory? httpClientFactoryParam = null, bool allowUseOfRateLimitedRegistryAPIs = true)
```

src/Shared/PackageManagers/ProjectManagerFactory.cs
morended: Addressed in the latest revision.

src/Shared/PackageManagers/ProjectManagerFactory.cs
morended: Addressed in the latest revision.

pmalmsten (approved):

PR #496, opened 2025-04-11
Add changes to support upstream to maven google repo

michelleqyun:
This PR adds the necessary changes to support Maven upstream to [Google Maven Repository](https://maven.google.com/web/index.html#). Some notes:

- Google Maven Repo has dynamic content in its website, so we use Puppeteer Sharp to stream and parse package metadata. 
- Caching is currently not enabled for requests through Puppeteer Sharp.
- Metadata requests to Maven Central Repository return contents of the .pom file, which are displayed in html on its website. We're unable to do that with Google Maven Repository, as it only contains a link to download the .pom file. To obtain metadata, we scrape whatever's available in html on the Google Maven Repository website. 
- If a requested package URL has an unrecognized/unsupported repository URL, the Maven Project Manager will default to no-op and returning null/empty values (whereas before it defaulted to the Maven Central Repository). But I'm considering changing this to default to Maven Central. Looking for feedback.

### DIFF
--- a/src/Shared/Model/Enums/MavenSupportedUpstream.cs
+++ b/src/Shared/Model/Enums/MavenSupportedUpstream.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.CST.OpenSource.Model.Enums;
+
+/// <summary>
+/// Maven upstream repositories supported by Terrapin.
+/// </summary>
+public enum MavenSupportedUpstream
+{
+    /// <summary>
+    /// https://repo1.maven.org/maven2
+    /// </summary>
+    MavenCentralRepository = 0,
+
+    /// <summary>
+    /// https://dl.google.com/android/maven2
+    /// </summary>
+    GoogleMavenRepository,
+}
+
+/// <summary>
+/// Extension methods for <see cref="MavenSupportedUpstream"/>.
+/// </summary>
+public static class MavenSupportedUpstreamExtensions
+{
+    /// <summary>
+    /// Gets the registry URI for the maven upstream repository.
+    /// </summary>
+    /// <param name="mavenUpstream">The <see cref="MavenSupportedUpstream"/>.</param>
+    public static string GetRepositoryUrl(this MavenSupportedUpstream mavenUpstream) => mavenUpstream switch
+    {
+        MavenSupportedUpstream.MavenCentralRepository => "https://repo1.maven.org/maven2",
+        MavenSupportedUpstream.GoogleMavenRepository => "https://maven.google.com/web/index.html#",
+        _ => throw new ArgumentOutOfRangeException(nameof(mavenUpstream), mavenUpstream, null),
+    };
+
+    /// <summary>
+    /// Returns the <see cref="MavenSupportedUpstream"/> associated with a given string.
+    /// </summary>
+    /// <param name="mavenUpstream"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentOutOfRangeException"></exception>
+    public static MavenSupportedUpstream GetMavenSupportedUpstream(this string mavenUpstream) => mavenUpstream switch
+    {
+        "MavenCentralRepository" => MavenSupportedUpstream.MavenCentralRepository,
+        "https://repo1.maven.org/maven2" => MavenSupportedUpstream.MavenCentralRepository,
+        "GoogleMavenRepository" => MavenSupportedUpstream.GoogleMavenRepository,
+        "https://dl.google.com/android/maven2" => MavenSupportedUpstream.GoogleMavenRepository,
+        _ => throw new ArgumentOutOfRangeException(nameof(mavenUpstream), mavenUpstream, null),
+    };
+}
--- a/src/Shared/PackageManagers/MavenProjectManager.cs
+++ b/src/Shared/PackageManagers/MavenProjectManager.cs
@@ -4,17 +4,22 @@ namespace Microsoft.CST.OpenSource.PackageManagers
 {
     using AngleSharp.Html.Parser;
     using Helpers;
+    using Microsoft.CodeAnalysis.Sarif;
     using Microsoft.CST.OpenSource.Contracts;
     using Microsoft.CST.OpenSource.Extensions;
     using Microsoft.CST.OpenSource.Model;
+    using Microsoft.CST.OpenSource.Model.Enums;
     using Microsoft.CST.OpenSource.PackageActions;
+    using Newtonsoft.Json;
     using PackageUrl;
+    using PuppeteerSharp;
     using System;
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
     using System.Net;
     using System.Net.Http;
+    using System.Text.RegularExpressions;
     using System.Threading.Tasks;
 
     public class MavenProjectManager : TypedManager<IManagerPackageVersionMetadata, MavenProjectManager.MavenArtifactType>
@@ -27,8 +32,8 @@ namespace Microsoft.CST.OpenSource.PackageManagers
 
         public override string ManagerType => Type;
 
-        public const string DEFAULT_MAVEN_ENDPOINT = "https://repo1.maven.org/maven2";
-        public string ENV_MAVEN_ENDPOINT { get; set; } = DEFAULT_MAVEN_ENDPOINT;
+        public const MavenSupportedUpstream DEFAULT_MAVEN_ENDPOINT = MavenSupportedUpstream.MavenCentralRepository;
+        public MavenSupportedUpstream ENV_MAVEN_ENDPOINT { get; set; } = DEFAULT_MAVEN_ENDPOINT;
 
         public MavenProjectManager(
             string directory,
@@ -42,27 +47,58 @@ namespace Microsoft.CST.OpenSource.PackageManagers
         /// <inheritdoc />
         public override async IAsyncEnumerable<ArtifactUri<MavenArtifactType>> GetArtifactDownloadUrisAsync(PackageURL purl, bool useCache = true)
         {
-            string? packageNamespace = Check.NotNull(nameof(purl.Namespace), purl?.Namespace).Replace('.', '/');
             string? packageName = Check.NotNull(nameof(purl.Name), purl?.Name);
             string? packageVersion = Check.NotNull(nameof(purl.Version), purl?.Version);
-            string feedUrl = (purl?.Qualifiers?["repository_url"] ?? ENV_MAVEN_ENDPOINT).EnsureTrailingSlash();
+            string feedUrl = purl?.Qualifiers?["repository_url"] ?? ENV_MAVEN_ENDPOINT.GetRepositoryUrl();
+            MavenSupportedUpstream upstream = feedUrl.GetMavenSupportedUpstream();
 
-            HttpClient httpClient = CreateHttpClient();
-            string baseUrl = $"{feedUrl}{packageNamespace}/{packageName}/{packageVersion}/";
-            string? html = await GetHttpStringCache(httpClient, baseUrl, useCache);
-            if (string.IsNullOrEmpty(html))
+            if (upstream == MavenSupportedUpstream.MavenCentralRepository)
             {
-                throw new InvalidOperationException();
+                string? packageNamespace = Check.NotNull(nameof(purl.Namespace), purl?.Namespace).Replace('.', '/');
+                var baseUrl = $"{upstream.GetRepositoryUrl().EnsureTrailingSlash()}{packageNamespace}/{packageName}/{packageVersion}/";
+
+                HttpClient httpClient = CreateHttpClient();
+                string? html = await GetHttpStringCache(httpClient, baseUrl, useCache);
+                if (string.IsNullOrEmpty(html))
+                {
+                    throw new InvalidOperationException();
+                }
+
+                HtmlParser parser = new();
+                AngleSharp.Html.Dom.IHtmlDocument document = await parser.ParseDocumentAsync(html);
+
+                foreach (string fileName in document.QuerySelectorAll("a").Select(link => link.GetAttribute("href").ToString()))
+                {
+                    if (fileName == "../") continue;
+
+                    MavenArtifactType artifactType = GetMavenArtifactType(fileName);
+                    yield return new ArtifactUri<MavenArtifactType>(artifactType, baseUrl + fileName);
+                }
             }
-
-            HtmlParser parser = new();
-            AngleSharp.Html.Dom.IHtmlDocument document = await parser.ParseDocumentAsync(html);
-            foreach (string fileName in document.QuerySelectorAll("a").Select(link => link.GetAttribute("href").ToString()))
+            else if (upstream == MavenSupportedUpstream.GoogleMavenRepository)
             {
-                if (fileName == "../") continue;
+                string packageNamespace = purl.Namespace;
+                var packageVersionUrl = $"{upstream.GetRepositoryUrl()}{packageNamespace}:{packageName}:{purl.Version}";
 
-                MavenArtifactType artifactType = GetMavenArtifactType(fileName);
-                yield return new ArtifactUri<MavenArtifactType>(artifactType, baseUrl + fileName);
+                // Google Maven Repository's webpage has dynamic content
+                var browserFetcher = new BrowserFetcher();
+                await browserFetcher.DownloadAsync();
+                var browser = await Puppeteer.LaunchAsync(new LaunchOptions { HeadlessMode = HeadlessMode.True });
+                var page = await browser.NewPageAsync();
+                await page.GoToAsync(packageVersionUrl, WaitUntilNavigation.DOMContentLoaded);
+
+                await page.WaitForSelectorAsync("tr");
+                var trElement = await page.XPathAsync("//tr[td[contains(text(), 'Artifact(s)')]]");
+                var anchors = await trElement.ElementAt(0).QuerySelectorAllAsync("a");
+
+                foreach (var anchor in anchors)
+                {
+                    var hrefValue = await (await anchor.GetPropertyAsync("href")).JsonValueAsync<string>();
+                    MavenArtifactType artifactType = GetMavenArtifactType(hrefValue);
+                    yield return new ArtifactUri<MavenArtifactType>(artifactType, hrefValue);
+                }
+
+                await browser.CloseAsync();
             }
         }
 
@@ -90,12 +126,12 @@ namespace Microsoft.CST.OpenSource.PackageManagers
             if (string.IsNullOrWhiteSpace(packageNamespace) || string.IsNullOrWhiteSpace(packageName) ||
                 string.IsNullOrWhiteSpace(packageVersion))
             {
-                Logger.Warn("Unable to download [{0} {1} {2}]. Both must be defined.", packageNamespace, packageName, packageVersion);
+                Logger.Warn("Unable to download [{0} {1} {2}]. All must be defined.", packageNamespace, packageName, packageVersion);
                 return downloadedPaths;
             }
 
             IEnumerable<ArtifactUri<MavenArtifactType>> artifacts = (await GetArtifactDownloadUrisAsync(purl, useCache: cached).ToListAsync())
-                .Where(artifact => artifact.Type is MavenArtifactType.Jar or MavenArtifactType.SourcesJar or MavenArtifactType.JavadocJar);
+                .Where(artifact => artifact.Type is MavenArtifactType.Jar or MavenArtifactType.SourcesJar or MavenArtifactType.JavadocJar or MavenArtifactType.Aar);
             foreach (ArtifactUri<MavenArtifactType> artifact in artifacts)
             {
                 try
@@ -142,12 +178,25 @@ namespace Microsoft.CST.OpenSource.PackageManagers
                 Logger.Trace("Provided PackageURL was null.");
                 return false;
             }
-            string packageNamespace = purl.Namespace.Replace('.', '/');
             string packageName = purl.Name;
-            string feedUrl = (purl?.Qualifiers?["repository_url"] ?? ENV_MAVEN_ENDPOINT).EnsureTrailingSlash();
+            string feedUrl = purl?.Qualifiers?["repository_url"] ?? ENV_MAVEN_ENDPOINT.GetRepositoryUrl();
+            MavenSupportedUpstream upstream = feedUrl.GetMavenSupportedUpstream();
+
             HttpClient httpClient = CreateHttpClient();
 
-            return await CheckHttpCacheForPackage(httpClient, $"{feedUrl}{packageNamespace}/{packageName}/maven-metadata.xml", useCache);
+            string packageRepositoryCheckUri = string.Empty;
+            if (upstream == MavenSupportedUpstream.MavenCentralRepository)
+            {
+                string packageNamespace = purl.Namespace.Replace('.', '/');
+                packageRepositoryCheckUri = $"{upstream.GetRepositoryUrl().EnsureTrailingSlash()}{packageNamespace}/{packageName}/maven-metadata.xml";
+            }
+            else if (upstream == MavenSupportedUpstream.GoogleMavenRepository)
+            {
+                string packageNamespace = purl.Namespace;
+                packageRepositoryCheckUri = $"{upstream.GetRepositoryUrl()}{packageNamespace}:{packageName}";
+            }
+
+            return await CheckHttpCacheForPackage(httpClient, packageRepositoryCheckUri, useCache);
         }
 
         /// <inheritdoc />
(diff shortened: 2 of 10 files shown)


pmalmsten:
@michelleqyun I see that Puppeteer Sharp is a web browser orchestrator. A few questions:
* Does Puppeteer Sharp require a browser to be separately installed on the system on which it runs? Would we need to ship a browser into our production servers and keep it up to date?
* Browser orchestration tends to be relatively slow. How performant is this approach?

scovetta:
I'd prefer we not ship a browser orchestrator if at all possible. From a quick look, it seems like the data is in predictable locations that we can just access directly:

* List of Versions: For com.google.cose:cose, `https://dl.google.com/android/maven2/com/google/cose/group-index.xml`.
* Metadata for a version: For com.google.cose:cose@20230908, `https://dl.google.com/android/maven2/com/google/cose/cose/20230908/cose-20230908.pom`
* Artifacts for a version: For com.google.cose:cose@20230908, `https://dl.google.com/android/maven2/com/google/cose/cose/20230908/artifact-metadata.json`
* Files: For com.google.cose:cose@20230908 / cose-20230908.jar (from the artifacts JSON), `https://dl.google.com/android/maven2/com/google/cose/cose/20230908/cose-20230908.jar`.

For the website, parsing is done via [gmaven-index.js](https://maven.google.com/web/js/gmaven-index.js) -- super basic, which just pulls data out of the JSON/POM/XML files, and lastModified from the HTTP Last-Modified header.

michelleqyun:
@scovetta I made your suggested changes. Let me know what you think!

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -27,8 +29,8 @@ public class MavenProjectManager : TypedManager<IManagerPackageVersionMetadata,
 
         public override string ManagerType => Type;
 
-        public const string DEFAULT_MAVEN_ENDPOINT = "https://repo1.maven.org/maven2";
-        public string ENV_MAVEN_ENDPOINT { get; set; } = DEFAULT_MAVEN_ENDPOINT;
+        public const MavenSupportedUpstream DEFAULT_MAVEN_ENDPOINT = MavenSupportedUpstream.MavenCentralRepository;
+        public MavenSupportedUpstream ENV_MAVEN_ENDPOINT { get; set; } = DEFAULT_MAVEN_ENDPOINT;
```
scovetta, line 33: I suspect this won't work with env variable overrides (https://github.com/microsoft/OSSGadget/blob/main/src/Shared/Helpers/EnvironmentHelper.cs#L17) -- we could update that function to account for non-string types, though.

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -160,33 +229,65 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
...
-                        var versionStr = htmlEntry.TrimEnd(htmlEntry[^1]);
-                        Logger.Debug("Identified {0} version {1}.", packageName, versionStr);
-                        versionList.Add(versionStr);
+                        string versions = packageElement.GetAttribute("versions");
+                        var versionsArray = versions.Split(',');
+
+                        foreach (string version in versionsArray)
+                        {
+                            versionList.Add(version.Trim());
+                        }
```
pmalmsten, line 290: Could we consolidate these implementations to read from `maven-metadata.xml`? It seems like most of the repositories listed at https://mvnrepository.com/repos publish that file. Google Maven repository and Maven Central also publish it: https://maven.google.com/android/arch/core/common/maven-metadata.xml

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -213,12 +314,25 @@ public override async Task<bool> PackageVersionExistsAsync(PackageURL purl, bool
...
+                    string packageNamespace = purl.Namespace.Replace('.', '/');
+                    packageRepositoryCheckUri = $"{upstream.GetRepositoryUrl().EnsureTrailingSlash()}{packageNamespace}/{packageName}/{purl.Version}/";
+                }
+                else if (upstream == MavenSupportedUpstream.GoogleMavenRepository)
+                {
+                    string packageNamespace = purl.Namespace;
+                    packageRepositoryCheckUri = $"{upstream.GetRepositoryUrl()}{packageNamespace}:{packageName}:{purl.Version}";
+                }
+
+                string? content = await GetHttpStringCache(httpClient, packageRepositoryCheckUri, useCache);
```
pmalmsten, line 335: Is there a more standard way we could check this? What if we sent a HEAD request for the `pom.xml` file that we expect to exist?

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -243,23 +357,40 @@ public override async Task<bool> PackageVersionExistsAsync(PackageURL purl, bool
...
+                }
+
+                if (upstream == MavenSupportedUpstream.MavenCentralRepository)
+                {
+                    return await GetHttpStringCache(httpClient, $"{upstream.GetRepositoryUrl().EnsureTrailingSlash()}{packageNamespace}/{packageName}/{version}/{packageName}-{version}.pom", useCache);
+                }
+                else if (upstream == MavenSupportedUpstream.GoogleMavenRepository)
+                {
+                    return await GetHttpStringCache(httpClient, $"{upstream.GetDownloadRepositoryUrl().EnsureTrailingSlash()}{packageNamespace}/{packageName}/{version}/{packageName}-{version}.pom", useCache);
                 }
```
pmalmsten, line 391: From my other comments on the extension methods, I think this could be consolidated into a single implementation.

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -285,13 +416,25 @@ public override async Task<bool> PackageVersionExistsAsync(PackageURL purl, bool
...
+            if (upstream == MavenSupportedUpstream.MavenCentralRepository)
+            {
+                string? packageNamespace = purl?.Namespace?.Replace('.', '/');
+                return new Uri($"{upstream.GetRepositoryUrl().EnsureTrailingSlash()}{packageNamespace}/{packageName}");
+            }
+            else if (upstream == MavenSupportedUpstream.GoogleMavenRepository)
+            {
+                string packageNamespace = purl.Namespace;
+                return new Uri($"{upstream.GetRepositoryUrl()}{packageNamespace}/{packageName}");
+            }
```
pmalmsten, line 434: Is there a way we could have a single implementation for this? The Google Maven repository must follow the same directory structure right?

I forget the intent for this particular method though - is this method intended to provide a URL that users can click on or something?

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -315,48 +459,67 @@ private static MavenArtifactType GetMavenArtifactType(string fileName)
...
-            // [1] - The date it was published
-            // [2] - The time it waas published
-            // [3] - The download count
-            string[] versionParts = versionContent.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
-            if (DateTime.TryParse($"{versionParts[1]} {versionParts[2]}", out DateTime publishDateTime))
+                // Split the version content into its individual parts.
+                // [0] - The version
+                // [1] - The date it was published
+                // [2] - The time it waas published
+                // [3] - The download count
```
pmalmsten, line 505: The last component is not a download count, it's the file size in bytes.
```suggestion
                // Split the version content into its individual parts.
                // [0] - The version
                // [1] - The date it was published
                // [2] - The time it was published
                // [3] - The file size in bytes
```

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -315,48 +459,67 @@ private static MavenArtifactType GetMavenArtifactType(string fileName)
...
+                string[] versionParts = versionContent.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
+                if (DateTime.TryParse($"{versionParts[1]} {versionParts[2]}", out DateTime publishDateTime))
+                {
+                    return publishDateTime;
+                }
+            }
```
pmalmsten, line 511: Since the directory index strategy is so common, should this be the default strategy? I understand that the Google Maven repo needs a custom strategy, but since the majority of Maven repos provide directory indexes, should we try this  for all other repositories than the Google repository?

pmalmsten, line 511: I've seen a couple places where we fetch the directory index. Should we extract this to a single implementation that we can reuse?

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -315,48 +459,67 @@ private static MavenArtifactType GetMavenArtifactType(string fileName)
...
-                return publishDateTime;
+                var baseUrl = $"{upstream.GetDownloadRepositoryUrl().EnsureTrailingSlash()}{packageNamespace}/{packageName}/{packageVersion}/{packageName}-{packageVersion}.pom";
+                HttpResponseMessage response = await httpClient.GetAsync(baseUrl);
+
+                // Try to get the "Last-Modified" header
+                if (response.Content.Headers.TryGetValues("Last-Modified", out var values))
+                {
+                    var lastModified = DateTime.Parse(values.First());
+                    return lastModified;
+                }
```
pmalmsten, line 522: For the handful of repositories that I have tested, if this header is provided (some don't provide it), then it seems accurate.

I wonder whether this should be a fallback approach for any repository type, if the directory index approach doesn't work? We could still have the Google Maven repository skip the directory index strategy since we know that won't work

src/Shared/Model/Enums/MavenSupportedUpstream.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <summary>
+    /// Gets the download URI for the maven upstream repository.
+    /// </summary>
+    /// <param name="mavenUpstream">The <see cref="MavenSupportUpstream"/>.</param>
+    public static string GetDownloadRepositoryUrl(this MavenSupportedUpstream mavenUpstream) => mavenUpstream switch
+    {
+        MavenSupportedUpstream.MavenCentralRepository => "https://repo1.maven.org/maven2",
+        MavenSupportedUpstream.GoogleMavenRepository => "https://dl.google.com/android/maven2",
+        _ => throw new ArgumentOutOfRangeException(nameof(mavenUpstream), mavenUpstream, null),
+    };
```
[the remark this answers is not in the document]
michelleqyun: The Google Maven repo frontend is different from the download URL to retrieve artifacts, eg: 

- https://maven.google.com/web/index.html#android.arch.core:common:1.1.1 [frontend]
- https://dl.google.com/android/maven2/android/arch/core/common/1.1.1/common-1.1.1.pom

I was using the frontend to do existence checks because it's a little simpler to check that the webpage for the package/package version exists rather than checking for artifact existence (the current Maven Central implementation also checks for the webpage). Especially for package existence checks with Google Maven, we would need to grab and parse the `group-index.xml` file (namespace-scoped) to see if the package name exists in the namespace. But if it makes more sense to change all the implementations to looking for artifacts, then I can do that too. What do you think?

src/Shared/Model/Enums/MavenSupportedUpstream.cs
```diff
@@ -0,0 +1,64 @@
...
+public static class MavenSupportedUpstreamExtensions
+{
+    /// <summary>
+    /// Gets the registry URI for the maven upstream repository.
+    /// </summary>
+    /// <param name="mavenUpstream">The <see cref="MavenSupportedUpstream"/>.</param>
+    public static string GetRepositoryUrl(this MavenSupportedUpstream mavenUpstream) => mavenUpstream switch
+    {
+        MavenSupportedUpstream.MavenCentralRepository => "https://repo1.maven.org/maven2",
+        MavenSupportedUpstream.GoogleMavenRepository => "https://maven.google.com/web/index.html#",
```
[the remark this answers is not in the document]
michelleqyun: I think it would be [https://dl.google.com/android/maven2](https://dl.google.com/android/maven2/android/arch/core/common/1.1.1/common-1.1.1.pom), but same idea.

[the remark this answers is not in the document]
michelleqyun: Nvm, I just saw your comment below that they redirect to the same link.

src/Shared/Model/Enums/MavenSupportedUpstream.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// Returns the <see cref="MavenSupportedUpstream"/> associated with a given string.
+    /// </summary>
+    /// <param name="mavenUpstream"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentOutOfRangeException"></exception>
+    public static MavenSupportedUpstream GetMavenSupportedUpstream(this string mavenUpstream) => mavenUpstream switch
+    {
+        "MavenCentralRepository" => MavenSupportedUpstream.MavenCentralRepository,
+        "https://repo1.maven.org/maven2" => MavenSupportedUpstream.MavenCentralRepository,
+        "GoogleMavenRepository" => MavenSupportedUpstream.GoogleMavenRepository,
```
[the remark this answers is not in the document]
michelleqyun: I was imagining a scenario where a user might have placeholder values, but it makes sense that that may end up being more confusing in the long run. I'll remove this portion.

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: With Maven Central, I noticed a couple cases where the version information in `maven-metadata.xml` did not match what was available on the frontend. For example, for the package https://repo1.maven.org/maven2/ant/ant/, the `maven-metadata.xml` only lists 12 versions, whereas 13 versions are available on the frontend. Should we still defer to the `maven-metadata.xml` as the source of truth for versioning?

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: The documentation for this method is: "Get the uri for the package home page (no version)". So it should yield something like: https://maven.google.com/web/index.html#android.arch.core:common (vs Maven Central: https://repo1.maven.org/maven2/ant/ant/).

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -42,27 +44,81 @@ public MavenProjectManager(
...
+                AngleSharp.Html.Dom.IHtmlDocument document = await parser.ParseDocumentAsync(html);
+
+                foreach (string fileName in document.QuerySelectorAll("a").Select(link => link.GetAttribute("href").ToString()))
+                {
+                    if (fileName == "../") continue;
+
+                    MavenArtifactType artifactType = GetMavenArtifactType(fileName);
+                    yield return new ArtifactUri<MavenArtifactType>(artifactType, baseUrl + fileName);
+                }
+            }
```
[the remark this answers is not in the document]
michelleqyun: I think that makes sense. If we're taking a default strategy, then should we add the concept of a `MavenSupportedUpstream.Unknown`?

src/Shared/Model/Enums/MavenSupportedUpstream.cs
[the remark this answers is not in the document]
pmalmsten: Yeah placeholder values would be fine for things like the OSSGadget CLI parser. As long as we can keep them away from databases 😅

src/Shared/PackageManagers/MavenProjectManager.cs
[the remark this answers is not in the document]
pmalmsten: To be honest @michelleqyun I wonder how much value we really get out of that enum. From the Terrapin end, we would largely be passing in URLs that we receive as the `repository_url` qualifier on package URLs from AA; would it be most straightforward to take the URL provided and use that directly without translating to an enum in between?

We could still have a list of well-known repository URLs and provide special casing for them as appropriate (like for Google Maven Repository).

src/Shared/PackageManagers/MavenProjectManager.cs
pmalmsten: That's very interesting. In that particular example, it looks like version `1.7.0` only has a pom.xml file (and no  .jars). Is `1.7.0` actually usable in a meaningful way?

src/Shared/PackageManagers/MavenProjectManager.cs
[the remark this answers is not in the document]
michelleqyun: I see your point. That might make more sense

src/Shared/Model/Enums/MavenSupportedUpstream.cs
```diff
@@ -0,0 +1,64 @@
...
+/// <summary>
+/// Extension methods for <see cref="MavenSupportedUpstream"/>.
+/// </summary>
+public static class MavenSupportedUpstreamExtensions
+{
+    /// <summary>
+    /// Gets the registry URI for the maven upstream repository.
+    /// </summary>
+    /// <param name="mavenUpstream">The <see cref="MavenSupportedUpstream"/>.</param>
+    public static string GetRepositoryUrl(this MavenSupportedUpstream mavenUpstream) => mavenUpstream switch
```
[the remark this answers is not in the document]
michelleqyun: Removed this method.

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: Yeah it seems like it's not actually usable. I ended up using the `maven-metadata.xml` for versioning. If that's not available, then trying to use the directory index.

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: I kept it separate for now because there's only one shared line that extracts the index. Any content parsing is unique based on what we're looking for.

michelleqyun: I made those changes so that we try the directory index strategy as a default, and then try to the "Last-Modified" header if that doesn't work. Let me know what you think.

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: Resolving this comment in favor of continuing this conversation in the other comment thread.

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: The only place I can find it being used is to build a SARIF Result.Location object for a given package URL (I think for logging purposes?). It doesn't seem that critical; we could also leave this method unimplemented.

michelleqyun: I removed this method. Feel free to reopen this thread if you have more thoughts.

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -47,78 +49,34 @@ public override async IAsyncEnumerable<ArtifactUri<MavenArtifactType>> GetArtifa
...
-                    {
-                        // .aar does not exist
-                    }
-
-                    // all package versions have a .pom file
-                    fileNamesList.Add($"{filePrefix}.pom");
+                    baseUrl += $"{purl.Name}-{purl.Version}";
+                    fileNamesList = await GetArtifactDownloadUris_FileProbeStrategyAsync(baseUrl, purl, useCache).ToListAsync();
                 }
+            }
```
pmalmsten, line 74: If these methods returned null when they did not succeed (to differentiate from succeeding but finding nothing), we could potentially simplify this code a bit using null coalescing operators like `??=`:

```c#
List<string>? fileNamesList = null;

// Most Maven repositories have web server directory indices (but not Google Maven).
// First attempt to obtain artifact download URIs via the index.
if (feedUrl != GOOGLE_MAVEN_ENDPOINT)
{
    fileNamesList = await GetArtifactDownloadUris_DirectoryIndexStrategyAsync(baseUrl, useCache);
}

fileNamesList ??= await GetArtifactDownloadUris_ArtifactMetadataStrategyAsync(baseUrl, useCache);
fileNamesList ??= await GetArtifactDownloadUris_FileProbeStrategyAsync($"{baseUrl}{purl.Name}-{purl.Version}", purl, useCache);
```

The child methods here would not be able to be `IAsyncEnumerable`. But are those methods really taking advantage of that interface? That interface is most useful when retrieving one more collection element might need to cause the caller to `await` again on e.g. a network fetch to fetch another page of results. But it looks like these helper methods don't do pagination like that, they get everything in one network fetch?

Not a particularly important point, I just like to look for ways we could simplify things.

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -231,36 +179,45 @@ public override async Task<IEnumerable<string>> EnumerateVersionsAsync(PackageUR
...
+                }
+                else
                 {
+                    // If maven-metadata.xml file is unavailable, try using the web server directory index.
+                    baseUrl = $"{feedUrl.EnsureTrailingSlash()}{packageNamespace}/{packageName}/";
+                    content = await GetHttpStringCache(httpClient, baseUrl, useCache);
+
+                    HtmlParser parser = new();
+                    AngleSharp.Html.Dom.IHtmlDocument document = await parser.ParseDocumentAsync(content);
+
```
pmalmsten, line 220: Could these benefit from the new helper methods?

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -501,28 +393,112 @@ private static MavenArtifactType GetMavenArtifactType(string fileName)
...
+                // [2] - The time it was published
+                // [3] - The file size in bytes
                 string[] versionParts = versionContent.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
                 if (DateTime.TryParse($"{versionParts[1]} {versionParts[2]}", out DateTime publishDateTime))
                 {
                     return publishDateTime;
                 }
             }
-            else if (upstream == MavenSupportedUpstream.GoogleMavenRepository)
+
```
pmalmsten, line 404: Could the above be combined with GetArtifactDownloadUris_DirectoryIndexStrategyAsync somehow?

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -501,28 +393,112 @@ private static MavenArtifactType GetMavenArtifactType(string fileName)
...
+                }
+            }
+            catch {}
+
+            foreach (var file in fileNameList)
             {
-                var baseUrl = $"{upstream.GetDownloadRepositoryUrl().EnsureTrailingSlash()}{packageNamespace}/{packageName}/{packageVersion}/{packageName}-{packageVersion}.pom";
-                HttpResponseMessage response = await httpClient.GetAsync(baseUrl);
+                yield return file;
+            }
```
pmalmsten, line 445: Nit: As I mentioned before, `IAsyncEnumerable` appears to be overkill for methods like these. As I understand, the primary value add of `IAsyncEnumerable` is that one could add statements between calls to `yield` that need to `await` on something to get more results; like `await`ing on a network call to get another page of results before yielding each item in each page.

When we already have everything in one shot, a simple `IEnumerable` might suffice. We might even be able to return the LINQ expression result directly, something like this:

```c#
return document.QuerySelectorAll("a")
  .Select(link => link.GetAttribute("href").ToString())
  .Where(fileName != "../")
  .Select(fileName => baseUrl + fileName);
```

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -501,28 +393,112 @@ private static MavenArtifactType GetMavenArtifactType(string fileName)
...
+        {
+            HttpClient httpClient = CreateHttpClient();
+            var fileNameList = new List<string>();
+
+            try
+            {
+                string? html = await GetHttpStringCache(httpClient, baseUrl, useCache);
+                if (string.IsNullOrEmpty(html))
+                {
+                    throw new InvalidOperationException();
```
pmalmsten, line 428: How could we make this easier to understand/debug if we end up throwing this exception?

Also - if we throw an exception, that means that we will not attempt any other of the strategies for finding directory contents. Is that intentional?

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -501,28 +393,112 @@ private static MavenArtifactType GetMavenArtifactType(string fileName)
...
+                HtmlParser parser = new();
+                AngleSharp.Html.Dom.IHtmlDocument document = await parser.ParseDocumentAsync(html);
+
+                foreach (string fileName in document.QuerySelectorAll("a").Select(link => link.GetAttribute("href").ToString()))
+                {
+                    if (fileName == "../") continue;
+                    fileNameList.Add(baseUrl + fileName);
+                }
+            }
+            catch {}
```
pmalmsten, line 440: How would we debug an unexpected exception from `GetHttpStringCache`?

src/Shared/PackageManagers/MavenProjectManager.cs
pmalmsten: Same note applies to the other strategies as well.

src/Shared/PackageManagers/MavenProjectManager.cs
pmalmsten: Same note applies to the other strategies as well.

src/Shared/PackageManagers/MavenProjectManager.cs
pmalmsten: Same note applies to the other strategies as well.

src/Shared/PackageManagers/MavenProjectManager.cs
pmalmsten: Hmm. I hear your point about the different methods selecting different things from the HTML. I'm not attached to consolidating these, but I can't help but wonder if we could benefit from having fewer places we need to think about error handling, debugging, etc. if we were to have one implementation of how to parse the directory index HTML that could fetch everything we might want. Something that could spit out like a list of file name & publish date time pairs, or a dictionary from file name to publish date time, etc.

pmalmsten: Looks great, thanks!

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: There's a catch block, so the other strategies should still be attempted. I can add some logging in the catch block so we can capture info on this exception.

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -47,78 +49,34 @@ public override async IAsyncEnumerable<ArtifactUri<MavenArtifactType>> GetArtifa
...
-                        fileNamesList.Add($"{filePrefix}.aar");
-                    }
-                    catch (HttpRequestException)
-                    {
-                        // .aar does not exist
-                    }
-
-                    // all package versions have a .pom file
-                    fileNamesList.Add($"{filePrefix}.pom");
+                    baseUrl += $"{purl.Name}-{purl.Version}";
```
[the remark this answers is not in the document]
michelleqyun: Moved it! Initially I wanted to avoid passing the purl to the helper method, but it makes more sense to revise the `baseUrl` in the helper method in case we add further helper methods down the line.

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: I added some additional exception handling and logging to `GetArtifactDownloadUris_FileProbeStrategyAsync` and `GetArtifactDownloadUris_ArtifactMetadataStrategyAsync`.

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: Okay, good to know. I changed the return type to `Task<List<string>>?` to simplify this and differentiate between unsuccessful vs. successful with no results.

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: Added the helper methods!

src/Shared/PackageManagers/MavenProjectManager.cs
michelleqyun: Yeah I can try the dictionary approach. FYI not all repositories that follow the directory index approach provide the published timestamp (e.g. [https://plugins.gradle.org/m2/Hoop/conductor-lifecycle/0.0.1/](https://plugins.gradle.org/m2/Hoop/conductor-lifecycle/0.0.1/)). But I can see the benefit of having one method for filename to publish datetime correlation.

src/Shared/Model/Enums/MavenArtifactType.cs
```diff
@@ -0,0 +1,46 @@
...
+        MavenArtifactType.Ear => ".ear",
+        MavenArtifactType.Jar => ".jar",
+        MavenArtifactType.JavadocJar => "-javadoc.jar",
+        MavenArtifactType.Pom => ".pom",
+        MavenArtifactType.Rar => ".rar",
+        MavenArtifactType.SourcesJar => "-sources.jar",
+        MavenArtifactType.TestsJar => "-tests.jar",
+        MavenArtifactType.TestSourcesJar => "-tests-sources.jar",
+        MavenArtifactType.War => ".war",
+        _ => string.Empty,
```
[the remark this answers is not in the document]
michelleqyun: I went with your suggestion because it helps simplify the code a bit, but there's still the concept of a "default" in the sense that if there's no type with a matching description, we will return `string.Empty`.

src/Shared/PackageManagers/MavenProjectManager.cs
```diff
@@ -285,81 +298,246 @@ public override async Task<bool> PackageVersionExistsAsync(PackageURL purl, bool
...
+
+                        // Trim any '/' from the version or file name
+                        publishedTimestampDict.Add(rowParts[0].Replace("/", string.Empty), publishedTimestamp);
+                    }
+                }
+            }
+            catch (Exception e)
             {
-                return publishDateTime;
+                Logger.Trace($"Directory index strategy for {purl} was unsuccessful: {e.Message}");
```
pmalmsten, line 536: Whenever we log exceptions, it's best practice to provide the exception object directly to the logger: https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#log-exceptions

This allows the logger to report the full stack trace in a well-formatted way.

pmalmsten: Same note applies for all of the exception handlers that log exceptions

michelleqyun: Added!

src/oss-tests/TestData/Maven/maven_core_1.0.0_alpha2.aar
```diff
@@ -0,0 +1,142 @@
+PK
```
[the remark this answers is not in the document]
michelleqyun: Yes, it's used the tests.

danfiedler-msft (approved):
lgtm

morended (approved):

PR #501, opened 2025-05-13
publish test results to pipeline run

danfiedler-msft:


danfiedler-msft:
This approach didn't work.

PR #502, opened 2025-05-14
experiment with migrating unit tests to xunit

danfiedler-msft:
We currently don't have test results in our pipeline runs. This makes it difficult to know which test(s) caused a failed run.

This PR:
- migrates all tests in the oss-tests project from MSTest to xunit + Microsoft.NET.Test.Sdk
- switches all test files to file-scoped namespaces and cleans up the usings (might as well since they all change due to the move to xunit)
- swaps AwesomeAssertions in for FluentAssertions due to the licensing changes of FluentAssertions


gfs (approved):
LGTM.

PR #530, opened 2026-03-25
Naomizheng/fix stale npm cache bug

naomizhengmsft:
NPM needs to fetch package entire metadata and stores it in the cache. Previously if the package we are trying to fetch exist in the cache, we would then assume it would have the version we need but the data in the cache could be stale especially if the new package was published within 30 minutes of the initial cache store so we should try to refetch the data or else we would return back null.


michelleqyun (approved):

morended (approved):

src/Shared/PackageManagers/NPMProjectManager.cs
```diff
@@ -599,8 +599,24 @@ is JsonElement.ArrayEnumerator enumeratorElement &&
...
+
+            
+            // If we got null from cache, it could mean:
+            // 1. Version doesn't exist
+            // 2. Cache is stale (version was published after cache was populated)
+            // Re-fetch and evict cache so next request gets fresh data.
+            if (publishTime == null && useCache)
+            {
+                DataCache.Remove($"{cacheUrl}/json");
+                jsonDoc = await GetJsonCache(client, cacheUrl, useCache: false);
```
pmalmsten, line 615: @naomizhengmsft  I'm a bit late to this PR, but this line states `useCache: false`. Isn't that redundant with `DataCache.Remove($"{cacheUrl}/json");` above? Would we be fine if this second call to GetJsonCache updated the cache again with a fresh value (since we just explicitly evicted any prior value there)?

PR #531, opened 2026-04-02
propogate ossgadget errors

naomizhengmsft:


gfs:
I think the pipeline errors here are possibly because this is a PR from a fork instead of a branch natively in OSSGadget?

danfiedler-msft (approved):

PR #534, opened 2026-06-04
Pin vulnerable transitive NuGet packages (SFI-ES5.2)

tzn07:
- Microsoft.Bcl.Memory -> 10.0.4
- System.Net.Http      -> 4.3.4
- System.Text.RegularExpressions -> 4.3.1

Added as direct <PackageReference> overrides in the consuming csprojs. CPM not introduced. Verified via dotnet restore/build/test against nuget.org (ADO feed unreachable from build env).

S360-Run-Id: 189d082f-67ad-401e-b203-788bd7c886cf
S360-KPI: SFI-ES5.2
S360-Skill: dependabot:dependency-update-orchestrator
S360-Arm: dedicated_skill

### DIFF
--- a/src/Shared.CLI/Shared.CLI.csproj
+++ b/src/Shared.CLI/Shared.CLI.csproj
@@ -96,6 +96,10 @@
     <PackageReference Include="SemanticVersioning" Version="3.0.0" />
     <PackageReference Include="System.Console" Version="4.3.1" />
     <PackageReference Include="Whois" Version="3.0.1" />
+    <!-- SFI-ES5.2: pin vulnerable transitive -->
+    <PackageReference Include="Microsoft.Bcl.Memory" Version="10.0.4" />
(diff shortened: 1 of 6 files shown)


JustinGOSSES (approved):

Ansul (commented):
Changes look to me from SFI compliance. I was wondering if it would be beneficial to move over the patched binaries to a store, if this is going to be continuous effort needing PR or is it just a one-time thing.

tzn07:
> Changes look to me from SFI compliance. I was wondering if it would be beneficial to move over the patched binaries to a store, if this is going to be continuous effort needing PR or is it just a one-time thing.

Yes it's for SFI compliance. We have https://github.com/microsoft/OSSGadget/pull/535 PR to automate the PR creation process for these type of PRs as a way to automate this.

morended (approved):

